PR #97, opened 2022-10-20
update:adding extra information regarding commits and test-export

Pradhyuman-sharma:


nehagup (changes requested):
@Pradhyuman-sharma this info needs to be added in the contribution guide. People might follow this command while server installation and then the test export would not work.

nehagup (approved):

PR #232, opened 2023-10-06
updated the title for version 1 and 2, fixes #966

Ritish134:
# Pull Request Template

## Description

Updated the title for version 1 and 2 in code-contribution.md file

Fixes #966


Sonichigo (changes requested):
This need to be done for all the pages with the same title tag. And the changes need to be specifically made for the V1, as the V2 are the current and active version so it's not needed!

Sonichigo:
Also, please sign the DCO's and fix the lint-codebase in the next commit.

PR #377, opened 2024-04-24
feat: remove sys from import

darkin424:
`import sys` is not required as well as how to install pytest wasn't mentioned anywhere in the guide, therefore have added `pip install keploy pytest`


Sonichigo (approved):
LGTM

PR #417, opened 2024-07-17
Solves GIFs seems grainy/pixelated in dark mode

starvader13:
## Description
The GIF gets grainy and pixelated in the dark mode, the solution is to change the background colour of the image to white.

Fixes https://github.com/keploy/keploy/issues/2020

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue).
- [X] This change requires a documentation update.

## Screenshot

Screenshot from 2024-07-17 19-14-06

Screenshot from 2024-07-17 19-14-38

- [X] My code follows the style guidelines of this project.

### DIFF
--- a/versioned_docs/version-2.0.0/concepts/what-are-keploy-features.md
+++ b/versioned_docs/version-2.0.0/concepts/what-are-keploy-features.md
@@ -58,7 +58,7 @@ within your CI environment without needing to connect to external services/depen
 No more data dumps, stubs, or mocks for dependencies like DBs, internal services, or third-party services like twilio,
 shopify, or stripe are required anymore. 💡
 
-<img src="/docs/img/mock-dependencies.png?raw=true" width="50%" alt="Mock Application Dependencies"/>
+<img src="/docs/img/mock-dependencies.png?raw=true" width="50%" alt="Mock Application Dependencies" style={{backgroundColor: '#EDEDED'}}/>
 
 [Idempotency](/concepts/general-glossary.md#2-idempotency) guarantees are also **not required** in the application.
 Multiple Reads after write operations can be replicated automatically too. 🔄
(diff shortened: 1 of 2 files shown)


starvader13:
@Swpn0neel Review Required!!

Swpn0neel:
> @Swpn0neel Review Required!!
> 
> Running yarn prettier --write generates a warning. From what I understand, this shouldn't cause any unexpected issues in the future. However, you might want to take a look at it.
> 
> Screenshot from 2024-07-17 19-26-03

If you've already have the setup to prettify Markdown files while saving, in your code editor; it won't be a problem.

But, this is an existing issue, if you want to fix it, you can raise a separate issue for it and I'll count it in under kROC!

Swpn0neel (changes requested):
But this defeats the purpose of Dark Mode. Have you tried altering the GIF by trying to upscale/denoise it?

starvader13:
> But this defeats the purpose of Dark Mode. Have you tried altering the GIF by trying to upscale/denoise it?

I have tried upscaling and denoising it but the graininess was not reducding, that's why I switched over to this solution. It would be great if you could suggest me something ?

starvader13:
> If you've already have the setup to prettify Markdown files while saving, in your code editor; it won't be a problem.
> 
> But, this is an existing issue, if you want to fix it, you can raise a separate issue for it and I'll count it in under kROC!

Yeah as I have gone through the docs, it seems the `jsxBracketSameLine` is deprecated and `bracketSameLine` is what getting used now. I will be creating an issue asap.

Swpn0neel:
> @Swpn0neel I have tried upscaling and denoising it but the graininess was not reducing, that's why I switched over to this solution. It would be great if you could suggest me something.

Ok, then we can go with this solution!

starvader13:
A similar solution could be applied to this one too.

image

starvader13:
@Swpn0neel Any update on this one? I have gone through docs and can not find any other `gif/image` which needs to be refactored.

PR #549, opened 2025-06-18
typo error , fixed ink to link

rajputdivyanshu81:
## What has changed?

Fixed the type error in glossary.js file - from ink to link


rajputdivyanshu81:
@amaan-bhati okay sir , i will remember this for the next , saw the codebase and see that so thought to fix and raise pr

PR #557, opened 2025-06-18
fix: correct typo in write-heading-ids script and add development scr…

devlohani99:
hanged "wnrite-heading-ids" → "write-heading-ids"
This was a simple typo that would have caused the script to fail

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,11 +14,7 @@
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
     "format": "prettier --write .",
-    "prepare": "husky install",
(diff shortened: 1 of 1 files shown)


amaan-bhati:
Hey @devlohani99 
We noticed that this PR doesn’t include any meaningful changes. In the future, please ensure that a PR is raised only if you’ve made substantial updates or solved a particular issue. This helps keep the process efficient and organized.
This being a part of an assignment in api fellowship would be counted since you were suppsoed to create a pr but wont be merged since this is not that relevant of a change. I’m sure you understand, and I appreciate your cooperation on this. We will be closing this PR for now.

PR #610, opened 2025-07-15
fix: the-broken-link-of-glossary

pratik-mahalle:
**Fix: Corrected broken link in glossary**
This pull request resolves an issue with a broken link within the documentation's glossary section.


amaan-bhati (commented):
I've addressed the changes, instead of using images in png/svg/jpg/jpegs try using webp format images since it helps us improve the performance and reduce the loading time of the webpage. Also, compress these images a little either by using cwebp or https://squoosh.app/
you can also use the following command to compress and convert images: cwebp -q 50 -lossless path/to/the/image.jpeg -o path/to/the/image.webp

PR #846, opened 2026-04-29
docs(static-dedup): document custom-dedup-fields → schema enums; note k8s-proxy auth WIP

khareyash05:
## What has changed?

Three related updates to the Keploy Cloud / Enterprise docs:

1. **Static Deduplication — `custom-dedup-fields` → OpenAPI enum values (new section).** Documents the dual role the `custom-dedup-fields` config now plays:
   - Sharpens the dedup signature (existing behaviour) so two requests with the same shape but different values for marked fields stay as distinct test cases.
   - **New:** during auto-replay schema generation, the distinct scalar values observed for each marked field are emitted as an `enum` array on the matching response (or request) property, so the values surface in the generated OpenAPI document and in the console's **Schemas** section. Includes the config example, a before/after schema snippet for `GetProducts200Response`, and the caveats: scalar leaves only, **50-distinct-value saturation cap** (drop-rather-than-truncate), and object-response shape (top-level JSON arrays don't match `response.X` paths).

2. **Static Deduplication trim (already in this branch).** Removes the prior, out-of-date `Custom dedup fields` and `Observe dedup effectiveness` (`/dedup/stats`) sections so the page reflects what's actually shipped today. The new section above replaces the removed `custom-dedup-fields` material with up-to-date content tied to the schema-enum behaviour.

3. **Kubernetes Proxy API — Authentication WIP notice (already in this branch).** Adds an `:::info In development` admonition above the Authentication section of `running-keploy/k8s-proxy-api.md` so readers don't try to wire the shared-token flow into automation while it's still moving.

## Type of change

- [x] New feature (non-breaking change which adds functionality) — documents the new schema-enum emission path for `custom-dedup-fields`.
- [x] Documentation update.

## How Has This Been Tested?

- Verified end-to-end against a self-hosted k8s-proxy + agent recording of a small Go service with three configured dedup fields (`response.id`, `response.version`, `response.product_id`).
- Stop-recording → auto-replay → schema generation produced an OpenAPI document with `enum` arrays on the marked properties exactly as documented in the new section. Sample observed values: `"product_id": { "enum": ["1","5","7","42","100","999","2024"], "type": "string" }`.
- Fetched the same schema through the api-server-compatible `/k8s-proxy/get/schema` endpoint and confirmed the `enum` arrays survive the read path that backs the console's Schemas page.
- Local `npm run build` / `npm run serve` checks still recommended before merge.

## Checklist

- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.

### DIFF
--- a/versioned_docs/version-4.0.0/keploy-cloud/static-deduplication.md
+++ b/versioned_docs/version-4.0.0/keploy-cloud/static-deduplication.md
@@ -2,7 +2,7 @@
 id: static-deduplication
 title: Static Deduplication
 sidebar_label: Static Deduplication
-description: Drop schema-identical live traffic at record time so only distinct request shapes become test cases. Configure per-endpoint value-aware dedup with custom fields.
+description: Drop schema-identical live traffic at record time so only distinct request shapes become test cases. Mark fields with custom-dedup-fields to keep value-distinct traffic and surface observed values as enums in the generated OpenAPI schema.
 tags:
   - deduplication
   - recording
@@ -17,6 +17,8 @@ keywords:
   - recording dedup
   - keploy enterprise
   - dedup stats
+  - openapi enum
+  - schema enum generation
 ---
 
 import ProductTier from '@site/src/components/ProductTier';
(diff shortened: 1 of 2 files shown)


officialasishkumar (approved):
LGTM

PR #865, opened 2026-05-28
docs(k8s-proxy-llm-workflow): document --cluster for local cloud replay

charankamarapu:
## Problem

The skill's documented `keploy cloud replay --app <ns.deployment> --branch-name <git branch>` command fails locally in the most common dev setup: `--app` is the K8s/proxy-mode path (`--app string ... K8s/proxy apps only.` in the CLI's own help), and it requires resolving an **active cluster** heartbeating to whatever api-server it's talking to. On a dev laptop pointing at a local api-server there typically is no active cluster, so the agent gets `no active clusters found` and bails — which is exactly what made the doc's clean-session validations fall back to the sandbox flow.

The cluster name lives on the app record (`origin.clusterName`), and `selectCluster` accepts a non-active SaaS cluster by name. The agent just doesn't know to fetch it or pass it.

## Fix

Three coordinated edits in `versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md`:

1. **Discovery — new step 3 (Cluster).** Agent now caches `origin.clusterName` via `getApp({appId: app_id})` at session start. (Note: `listApps` does **not** return this — `getApp` is required.)
2. **Routine B / Phase B4 — Validate.** Documents the local form: `keploy cloud replay --app <ns.deployment> --cluster "<cluster_name>" --branch-name <git branch> -c "<dev run command>" --container-name <app container> --disableReportUpload=false`. CI / active-cluster runs drop the local flags.
3. **Routine A — re-validate command + Prompt B table B4 row.** Mirror the same flag set for consistency.

All three values (`app_id`, `branch_id`, `cluster_name`) are sticky for the session; the original `Both values are sticky` line becomes `All three values are sticky`.

## Why `--disableReportUpload=false`

Documented in the comment: the CLI silently sets `--disableReportUpload=true` for OAuth sessions, which hides the `/tr` report locally. Forcing `false` makes B5's run-report URL valid.

## Validation

Clean headless Claude session at a zero-memory path executed Routine B end-to-end against an unmodified version of this skill (post-edit), no improvisation: `getApp` → `--cluster "new app"` → `cloud replay` → 6/6 PASSED on a branch (run `77050487-748e-4f01-ac22-accac70da92b`). Before these edits, the same agent fell back to the MCP `create_test_suite` + `record_sandbox_test` sandbox flow because the documented B4 command errored out locally.

## Related (required together for the local path to work end-to-end)

- keploy/enterprise#2043 — `cloud replay` resolves compose-gen resources from any test set in the release (not just `TestSets[0]`).
- keploy/api-server#1697 — tab-safe YAML in `EncodeMockBundle` (round-trip fix for PostgresV3 `sqlNormalized` mocks).

### DIFF
--- a/versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
+++ b/versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
@@ -190,6 +190,8 @@ The contract changed on purpose; the test's recorded baseline is stale. Read `os
 
 **Case 2b loop — follow exactly, do not improvise:**
 
+After step 1's `update_mock` lands and the re-replay is still red, your repertoire collapses to **exactly two moves**: another `update_mock` (step 2) or `delete_recording` + re-record (step 3). Anything else — editing `keploy/<test_set_id>/mocks.yaml`, passing `--useLocalMock`, `--disableMockUpload`, or `--useLocalTests` to `keploy cloud replay`, comparing local YAML to spot "what changed", restarting the agent — is **off the menu**. Those flags exist for OSS-level proxy debugging; they appear to work but pin a divergent state to your laptop while the Keploy branch the team replays against stays broken. The contract is "the cloud branch is the only source of truth"; respect it or get out of Routine A.
+
 1. **First edit.** `getMock` → mutate the canonical YAML to reflect the new contract → `update_mock`. Re-run replay.
 2. **Second edit (only if step 1 still red on the same mock for the same reason).** Re-read `oss_report.mock_mismatches` for the new run; the diff between `expected_mocks` and `actual_mocks` should now be tighter. `getMock` again (the server may have rewritten derived fields), mutate, `update_mock`. Re-run replay.
 3. **Fallback (if step 2 is still red).** Recorded baseline is too far gone to patch piecemeal — choose between a **whole-set** drop and a **scoped** drop based on how many test cases are failing:
@@ -206,25 +208,33 @@ The contract changed on purpose; the test's recorded baseline is stale. Read `os
    ```
 
    **3b — Scoped re-record (only one or a few cases in the set are failing):**
+
+   > **Availability**: 3b requires an api-server that advertises `test_case_ids` on the `delete_recording` MCP tool. Run `listMocks`/`getApp`-style discovery once to confirm the input schema lists `test_case_ids`; if it doesn't, fall back to 3a until the deployment includes the scoped-delete change.
    ```
    delete_recording({                                            # tombstones JUST those cases;
      app_id, test_set_id, branch_id,                             # the rest of the set + its
-     test_case_ids: [<id-1>, <id-2>, ...]                        # mocks stay intact
-   })
+     test_case_ids: [<name-1>, <name-2>, ...]                    # mocks stay intact.
+   })                                                            # Names are the same friendly
+                                                                 # identifiers used elsewhere
+                                                                 # (e.g. "get-api-orders-1").
    keploy record -c "<dev run command>" --sync                   # capture only the dropped flows
    # drive curls for ONLY the test cases you tombstoned — use the
    # recorded request body of each as the curl shape
    keploy upload test-set \
      --app <ns.deployment> --branch <git branch> \
-     --test-set keploy/test-set-N --name <fresh-distinct-name>   # NEW name, distinct from
-   # re-run keploy cloud replay                                   # any existing set on this app
+     --test-set keploy/test-set-N \
+     --name <original-set-name>--rerec-<YYYYMMDDHHMM>            # convention: original + rerec suffix
+   # re-run keploy cloud replay                                   # so re-records cluster with the
+                                                                 # set they refresh
    ```
 
-   In 3b the branch ends with two coexisting test-sets: the original (minus the tombstoned cases) and the new small one with the replacements — both contribute to the next replay. Mint a fresh `--name`; the server rejects duplicates with `test set "X" already exists for this app`.
+   In 3b the branch ends with two coexisting test-sets: the original (minus the tombstoned cases) and the new small one with the replacements — both contribute to the next replay. Mint a fresh `--name` using the suffix convention above (or `<original>--rerec-<short-git-sha>` if a deterministic name is preferable); the server rejects duplicates with `test set "X" already exists for this app`, and the suffix keeps the recordings page self-grouping (original + its re-records sort together).
 
    Pick 3a when ≥ ~75% of the set's cases fail, 3b otherwise. Defaulting to 3a when only one case is failing destroys unrelated passing tests for no reason.
 
-**Do NOT inspect or edit `keploy/<test_set_id>/mocks.yaml` on the local filesystem between attempts.** `keploy cloud replay` re-downloads mocks from the Keploy branch on every run; any local edit is silently overwritten before the test runs. All mock changes go through `update_mock`. If you find yourself reading or grepping local mock files to "verify the change propagated", the answer is: it does, by construction — the local file is a per-run snapshot, not state you can edit.
+**Do NOT inspect or edit `keploy/<test_set_id>/mocks.yaml` on the local filesystem.** `keploy cloud replay` re-downloads mocks from the Keploy branch on every run; any local edit is silently overwritten before the next replay. All mock changes go through `update_mock`. If a local-edit + `--useLocalMock`-style workaround tempts you because cloud round-trip changed the shape of a value (e.g. a Postgres NUMERIC `price` came back as `{int: []}` instead of `{int: "1250", exp: -2}`), that is a step-2 retry signal — go to step 2 of the loop with a corrected payload, or to step 3 if step 2 also fails. Do not paper over the round-trip locally; the next replay on a CI runner or another laptop will undo your fix and the team-visible branch state will diverge from what you tested.
+
+**Do NOT pass `--useLocalMock`, `--useLocalTests`, or `--disableMockUpload` to `keploy cloud replay`.** Those flags belong to OSS-level proxy debugging and pin a laptop-local divergent state — the Keploy branch the rest of the team replays against stays broken regardless of what your local run reports. Every replay you trigger in Routine A must round-trip through the cloud branch; otherwise you are not in Routine A any more.
 
 **Do NOT ask the dev which path to take.** Routine A is autonomous. Announce the file:line you intend to change in plain text (so the dev can interrupt if they object), then proceed. Do not call `AskUserQuestion`, do not offer numbered choices, do not pause for confirmation. If two repair paths look equally valid, pick the one the skill recommends (Case 2a noise > body update, Case 2b update_mock > delete + re-record) and proceed. The dev is reviewing the streamed transcript and will Ctrl-C if you're wrong; that's the contract.
 
@@ -241,7 +251,7 @@ keploy cloud replay --app <ns.deployment> --cluster "<cluster_name>" --branch-na
 
 `--cluster` resolves the proxy app's identity without requiring an active heartbeat (use the `cluster_name` cached in Discovery). `-c` + `--container-name` start the app locally; omit them in CI / active-cluster runs and let the in-cluster agent run the deployment. `--disableReportUpload=false` makes the `/tr` report persist locally (the CLI silently sets it to `true` for OAuth sessions otherwise).
 
-If still failing, re-enter Phase A2 with the new `test_run_id`. If passing, proceed to A5. Cap retry attempts at 3—if it's still red, the failures are likely a keploy-side proxy issue (your fixes aren't taking effect). Report the residual failures honestly with the `test_run_id` and the run-report URL so the dev can file a keploy bug, then stop.
+If still failing, re-enter Phase A2 with the new `test_run_id`. If passing, proceed to A5. Cap retry attempts at 3 cloud-replay runs total across the Case 2 loop (i.e. step 1's replay + step 2's replay + step 3's replay = 3 — this is independent of the per-step retry budget within Case 2b, which caps `update_mock` attempts at 2 before forcing the 3a/3b fallback). If it's still red after 3 cloud-replay runs, the failures are likely a keploy-side proxy issue (your fixes aren't taking effect). Report the residual failures honestly with the `test_run_id` and the run-report URL so the dev can file a keploy bug, then stop.
 
 ### Phase A5—Report (exact format)
 
--- a/versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
+++ b/versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
@@ -237,7 +237,7 @@ After step 1's `update_mock` lands and the re-replay is still red, your repertoi
    >
    > Combined example: `scenario-4-v8-baseline-4-cases--rerec-202605281430Z`. Two agents running in different timezones at the same wall-clock minute produce the same name — preserving the "original + its re-records sort together" intent for cross-timezone CI integrations.
    >
-   > **Deterministic alternative.** If the original-set-name isn't available at name-mint time, use `rerec-<short-git-sha>-<utc-timestamp>` where `<short-git-sha>` is the first 7 characters of `git rev-parse HEAD`. This form drops the original-set prefix, so re-records won't sort-group with the original on the recordings page — only use it when the prefix is genuinely unavailable.
+   > **Deterministic alternative.** If the original-set-name isn't available at name-mint time, use `rerec-<short-git-sha>-<utc-timestamp>` where `<short-git-sha>` is the first 7 characters of `git rev-parse HEAD`. The alternative begins with the literal `rerec` (NO leading `-` or `--` — the double-dash in the primary form is the slug/suffix boundary marker, which this form doesn't have because there's no slug part). This form drops the original-set prefix, so re-records won't sort-group with the original on the recordings page — only use it when the prefix is genuinely unavailable.
 
    Pick 3a when ≥ ~75% of the set's cases fail, 3b otherwise. Defaulting to 3a when only one case is failing destroys unrelated passing tests for no reason.
 
--- a/versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
+++ b/versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
@@ -192,7 +192,37 @@ The contract changed on purpose; the test's recorded baseline is stale. Read `os
 
 1. **First edit.** `getMock` → mutate the canonical YAML to reflect the new contract → `update_mock`. Re-run replay.
 2. **Second edit (only if step 1 still red on the same mock for the same reason).** Re-read `oss_report.mock_mismatches` for the new run; the diff between `expected_mocks` and `actual_mocks` should now be tighter. `getMock` again (the server may have rewritten derived fields), mutate, `update_mock`. Re-run replay.
-3. **Fallback (if step 2 is still red).** The recorded baseline is too far gone to patch piecemeal. `delete_recording` on the affected test set, then re-capture from scratch via Routine B's flow: `keploy record -c "<dev run command>" --sync` against the current behavior → curl the affected paths → `keploy upload test-set --app <ns.deployment> --branch <git branch> --test-set keploy/test-set-N`. Re-run replay.
+3. **Fallback (if step 2 is still red).** Recorded baseline is too far gone to patch piecemeal — choose between a **whole-set** drop and a **scoped** drop based on how many test cases are failing:
+
+   **3a — Whole-set re-record (most / all cases in the set are failing):**
+   ```
+   delete_recording({app_id, test_set_id, branch_id})           # drops the entire set
+   keploy record -c "<dev run command>" --sync                   # captures all flows
+   # drive curls covering the same surface the original set covered
+   keploy upload test-set \
+     --app <ns.deployment> --branch <git branch> \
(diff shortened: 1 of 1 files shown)


versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
```diff
@@ -187,7 +188,45 @@ The contract changed on purpose; the test's recorded baseline is stale. Read `os
...
+   # re-run keploy cloud replay
+   ```
+
+   **3b — Scoped re-record (only one or a few cases in the set are failing):**
+   ```
+   delete_recording({                                            # tombstones JUST those cases;
+     app_id, test_set_id, branch_id,                             # the rest of the set + its
+     test_case_ids: [<id-1>, <id-2>, ...]                        # mocks stay intact
+   })
+   keploy record -c "<dev run command>" --sync                   # capture only the dropped flows
```
charankamarapu, line 222: **[cross-PR consistency — HIGH] 3b's `test_case_ids` param is gated on a server feature that isn't merged yet**

The 3b code block calls `delete_recording({app_id, test_set_id, branch_id, test_case_ids: [...]})`. The scoped-delete API surface that accepts `test_case_ids` was reverted from api-server PR #1697 (commit `2a11b4515 revert(apiv1,mcp): drop scope-aware delete_recording`), pending the enterprise CLI cache-reconciliation fix (PR #2047). Until both land + deploy, an agent following 3b will get back an error from `delete_recording` because the MCP tool doesn't advertise `test_case_ids` yet — and there's no fallback in the doc telling the agent what to do.

Two ways to fix:

- **Phase the doc** — add a tiny banner above 3b: "3b requires api-server ≥ <version> with scoped `delete_recording`. If the MCP tool's input schema doesn't include `test_case_ids`, fall back to 3a until the cluster updates."
- **Merge 1697 + 2047 first, then merge this** — cleanest, but couples three PRs across three repos.

Without one of these, the doc and the runtime contract diverge silently — agents reading 3b will pick the scoped path and hit a wall.

charankamarapu, line 222: **[clarity — MEDIUM] `test_case_ids: [<id-1>, ...]` overloads "id" against the rest of the doc**

Elsewhere the doc consistently teaches that `testCaseId` is the **friendly name** from the recording yaml (line 184 of the file: `updateTestCase` ... 'testCaseId is the friendly recording name, e.g. test-4'). Here in 3b the placeholder `<id-1>` reads like a UUID. An agent that reads this section literally may try to send the branch-overlay entity UUID (which it doesn't have) and get a 404.

```suggestion
   delete_recording({                                            # tombstones JUST those cases;
     app_id, test_set_id, branch_id,                             # the rest of the set + its
     test_case_ids: [<name-1>, <name-2>, ...]                    # mocks stay intact.
   })                                                            # Names are the same friendly
                                                                 # identifiers used elsewhere
                                                                 # (e.g. "get-api-orders-1").
```

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
```diff
@@ -187,7 +188,45 @@ The contract changed on purpose; the test's recorded baseline is stale. Read `os
...
+   # recorded request body of each as the curl shape
+   keploy upload test-set \
+     --app <ns.deployment> --branch <git branch> \
+     --test-set keploy/test-set-N --name <fresh-distinct-name>   # NEW name, distinct from
+   # re-run keploy cloud replay                                   # any existing set on this app
+   ```
+
+   In 3b the branch ends with two coexisting test-sets: the original (minus the tombstoned cases) and the new small one with the replacements — both contribute to the next replay. Mint a fresh `--name`; the server rejects duplicates with `test set "X" already exists for this app`.
+
+   Pick 3a when ≥ ~75% of the set's cases fail, 3b otherwise. Defaulting to 3a when only one case is failing destroys unrelated passing tests for no reason.
```
charankamarapu, line 244: **[guidance — MEDIUM] "Mint a fresh `--name`" without a convention causes naming sprawl**

The instruction says: "Mint a fresh `--name`; the server rejects duplicates with `test set "X" already exists for this app`." An agent following this literally will pick something arbitrary (`get-orders-rerec-v7` per the v7 run we just did) — and over many fallbacks, the app accumulates a long tail of one-off test-sets that humans later have to triage.

Suggest a stable convention agents can derive deterministically. Two reasonable shapes:

```
<original-set-name>--rerec-<YYYYMMDDHHMM>           # human-readable, sortable
<original-set-name>--rerec-<short-git-sha>           # deterministic if the agent has it
```

Without one, every agent picks its own; with one, the recordings page becomes self-grouping (original set + its re-records cluster together).

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
```diff
@@ -196,9 +235,12 @@ Multiple failing test cases can land in different cases—handle each independen
 After every Case-1 (app code edit) or Case-2 (test data edit) fix, run via Bash:
 
 ```bash
-keploy cloud replay --app <ns.deployment> --branch-name <git branch>
+keploy cloud replay --app <ns.deployment> --cluster "<cluster_name>" --branch-name <git branch> \
+  -c "<dev run command>" --container-name <app container> --disableReportUpload=false
 ```
```
charankamarapu, line 262: **[clarity — LOW] "Cap retry attempts at 3" (line 248) can be misread as the Case 2b loop budget**

Phase A4's pre-existing "Cap retry attempts at 3" line (just below this hunk) predates the Case 2b 1→2→fallback loop this PR adds. The two budgets cover different things:

- Case 2b loop: **2** `update_mock` attempts on the same mock before fallback.
- A4 cap: **3** full `keploy cloud replay` re-runs (step 1 replay + step 2 replay + step 3 replay) before declaring "likely keploy-side bug".

An agent reading them together might either (a) stop after 3 mock-edit attempts thinking that's the cap, or (b) re-enter the loop 3 times. Both wrong.

Worth a one-line clarification on the A4 cap line that the two budgets are independent.

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
charankamarapu: Fixed in `0204c5acf079a2414386e337df991061f2245ef8`. Added an Availability banner above the 3b block instructing the agent to confirm `test_case_ids` is on the `delete_recording` input schema and fall back to 3a if the api-server is older — agent makes progress on either deployment instead of dead-ending. Resolving.

charankamarapu: Fixed in `0204c5acf079a2414386e337df991061f2245ef8`. Applied the suggestion — placeholders are now `<name-1>`/`<name-2>` with an inline comment clarifying the values are the same friendly recording names used elsewhere (e.g. `get-api-orders-1`). Resolving.

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
charankamarapu: Fixed in `0204c5acf079a2414386e337df991061f2245ef8`. Convention is now `<original-set-name>--rerec-<YYYYMMDDHHMM>` (or `<short-git-sha>` as a deterministic alternative). Documented inline + in the surrounding prose so the recordings page becomes self-grouping (original + its re-records sort together) instead of an ad-hoc long tail. Resolving.

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
charankamarapu: Fixed in `0204c5acf079a2414386e337df991061f2245ef8`. The A4 cap line now spells out the relationship: 3 = total cloud-replay runs across the Case 2 loop (step 1 replay + step 2 replay + step 3 replay); 2 = `update_mock` attempts within step 2 before forcing 3a/3b fallback. Two budgets, named distinctly. Resolving.

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
```diff
@@ -187,7 +188,55 @@ The contract changed on purpose; the test's recorded baseline is stale. Read `os
...
+   # drive curls covering the same surface the original set covered
+   keploy upload test-set \
+     --app <ns.deployment> --branch <git branch> \
+     --test-set keploy/test-set-N --name <fresh-descriptive-name>
+   # re-run keploy cloud replay
+   ```
+
+   **3b — Scoped re-record (only one or a few cases in the set are failing):**
+
+   > **Availability**: 3b requires an api-server that advertises `test_case_ids` on the `delete_recording` MCP tool. Run `listMocks`/`getApp`-style discovery once to confirm the input schema lists `test_case_ids`; if it doesn't, fall back to 3a until the deployment includes the scoped-delete change.
```
charankamarapu: **[r2 — clarity — MEDIUM] "listMocks/getApp-style discovery" sends the agent to the wrong introspection method**

`listMocks` and `getApp` are domain endpoints — neither returns the MCP tool input schema. The canonical way for an MCP client to introspect a tool's input shape is via the protocol-level `tools/list` JSON-RPC method (returns each tool's `inputSchema`). An agent reading this banner literally will call `listMocks` or `getApp`, find nothing about `test_case_ids`, and either give up or guess.

```suggestion
   > **Availability**: 3b requires an api-server that advertises `test_case_ids` on the `delete_recording` MCP tool. To check, look at the `delete_recording` tool's `inputSchema.properties` in the MCP server's `tools/list` response — if `test_case_ids` (type: array of string) is listed, 3b is available. If it isn't, fall back to 3a until the deployment includes the scoped-delete change.
```

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
```diff
@@ -187,7 +188,55 @@ The contract changed on purpose; the test's recorded baseline is stale. Read `os
...
+   })                                                            # Names are the same friendly
+                                                                 # identifiers used elsewhere
+                                                                 # (e.g. "get-api-orders-1").
+   keploy record -c "<dev run command>" --sync                   # capture only the dropped flows
+   # drive curls for ONLY the test cases you tombstoned — use the
+   # recorded request body of each as the curl shape
+   keploy upload test-set \
+     --app <ns.deployment> --branch <git branch> \
+     --test-set keploy/test-set-N \
+     --name <original-set-name>--rerec-<YYYYMMDDHHMM>            # convention: original + rerec suffix
```
charankamarapu: **[r2 — guidance — MEDIUM] `<original-set-name>--rerec-<…>` collides with the upload endpoint's name validation when the original has spaces/parens/special chars**

Test sets are routinely named things like `scenario-4 v8 baseline (4 cases, OLD 9-col mocks)`. Concatenating `--rerec-<YYYYMMDDHHMM>` produces:

> `scenario-4 v8 baseline (4 cases, OLD 9-col mocks)--rerec-202605281430`

The api-server's `UploadTestSet` path-validates names — spaces and parens may or may not be accepted (worth verifying against `validateNameLike` in the api-server). If they're rejected, the entire 3b flow stops at `keploy upload test-set --name <bad>`. Two safer shapes:

```suggestion
     --name <slug(original-set-name)>--rerec-<YYYYMMDDHHMMZ>      # slug = lowercase, [a-z0-9-] only
```

OR drop the prefix and use a stable hash-of-original-id:

```
     --name rerec-<short-hash-of-original-set-id>-<YYYYMMDDHHMMZ>
```

Either avoids the validation footgun. The first preserves human-readable grouping; the second is more robust.

charankamarapu: **[r2 — clarity — LOW] `<YYYYMMDDHHMM>` doesn't specify timezone**

Two agents on different machines (CI runner in UTC vs dev laptop in IST) running the same `/loop` at the same wall-clock minute will mint different names. The purpose of the convention — "recordings page self-groups: original + its re-records sort together" — is partly defeated because two re-records of the same original from different timezones end up sorting separately.

Spell out UTC explicitly (I'd write it as `<YYYYMMDDHHMMZ>` to make the timezone literal in the name itself — agents who can't read the surrounding doc still emit UTC):

```suggestion
     --name <original-set-name>--rerec-<YYYYMMDDHHMMZ>            # YYYYMMDDHHMM in UTC; the trailing Z is part of the name
```

Inexpensive — one character — and removes a deterministic-naming gotcha for any future agent or CI integration.

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
charankamarapu: Fixed in `fc632665`. The banner now routes agents to MCP `tools/list` (the protocol-level introspection that returns each tool's `inputSchema`) and explicitly tells them to look for `test_case_ids` in `inputSchema.properties` on the `delete_recording` tool. No more pointing at `listMocks`/`getApp` which are domain endpoints and don't carry tool shape.

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
charankamarapu: Fixed in `fc632665`. The convention is now `<slug(original-set-name)>--rerec-<…>` where the slug rule is documented inline: lowercase, everything outside `[a-z0-9-]` collapsed to `-`. Worked example: `Scenario 4 v8 baseline (4 cases)` → `scenario-4-v8-baseline-4-cases`. Avoids the validator footgun for any original name carrying spaces/parens/other special chars. Also added the deterministic alternative `rerec-<short-git-sha>-<YYYYMMDDHHMMZ>` for callers who prefer hash-prefixing.

charankamarapu: Fixed in `fc632665`. Suffix is now `<YYYYMMDDHHMMZ>` — the trailing literal `Z` makes UTC visible in the name itself, and the new naming-convention note spells out "timestamp is **UTC**" explicitly. Two agents in different timezones at the same wall-clock minute now produce the same name, preserving the original/re-record grouping the convention is meant to enforce.

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
```diff
@@ -187,7 +188,64 @@ The contract changed on purpose; the test's recorded baseline is stale. Read `os
...
+     --app <ns.deployment> --branch <git branch> \
+     --test-set keploy/test-set-N \
+     --name <slug-part>--rerec-<utc-timestamp-part>              # build per "Naming convention details" below; do NOT post-process the result
+   # re-run keploy cloud replay                                   # so re-records cluster with the
+                                                                 # set they refresh
+   ```
+
+   In 3b the branch ends with two coexisting test-sets: the original (minus the tombstoned cases) and the new small one with the replacements — both contribute to the next replay. The server rejects duplicate names with `test set "X" already exists for this app`, so the convention below mints a unique `--name` and the shared prefix keeps the recordings page self-grouping (original + its re-records sort together).
+
+   > **Naming convention details.** Build the name in two parts and concatenate them verbatim — do NOT post-process the whole result.
```
charankamarapu, line 235: **[r3 — clarity — MEDIUM] Slug definition has three ambiguities; different agents will mint different names for the same source**

The round-2 wording closed the special-chars-fail-validation footgun but introduced three new ambiguities. Each is small on its own; together two agents implementing the rule literally can produce three different `--name` values for the same original.

**(1) Trailing literal `Z` collides with the slug charset `[a-z0-9-]`.** An agent that interprets the whole rule as "output must match `[a-z0-9-]`" lowercases the `Z` to `z`, breaking the UTC marker intent. Two agents — one strict-slugger, one literal-suffix-preserver — produce `...--rerec-202605281430z` vs `...--rerec-202605281430Z`.

**(2) Slug rule is under-specified vs the worked example.** Stated rule: "lowercased with everything outside `[a-z0-9-]` collapsed to `-`". Worked example: `Scenario 4 v8 baseline (4 cases)` → `scenario-4-v8-baseline-4-cases`.

Trace the stated rule literally → `scenario-4-v8-baseline--4-cases-` (double dash from `space(`, trailing dash from `)`). The example silently applies two more rules — "collapse runs of `-` to one" and "trim leading/trailing `-`". Neither stated.

**(3) `<slug(original-set-name)>` reads as a function-call placeholder.** An LLM agent following literally may search for a `slug` tool/CLI subcommand or fall back to a slugify library (python-slugify, django.utils.text.slugify, lodash.kebabCase). These produce divergent outputs for the same input.

**Fix:** make the rules complete and the placeholder declarative. Two-part declarative spec with explicit "do NOT post-process the whole result":
- Slug part: `<original-set-name>`, lowercased, runs of non-`[a-z0-9]` collapsed to one `-`, leading/trailing `-` trimmed. Worked example.
- Suffix part: literal `--rerec-<YYYYMMDDHHMMZ>` where `Z` is uppercase part of the literal — do NOT lowercase or slug it.
- Spell out `<short-git-sha>` = first 7 chars of `git rev-parse HEAD` for the deterministic alternative.

charankamarapu: Fixed in `545be5e99492bc2abef732f93b485429887f46d0`. Replaced the round-2 wording with a two-part declarative spec that closes all three ambiguities:

1. **Slug part**: explicitly "lowercase, replace every **maximal run** of characters outside `[a-z0-9]` (spaces, parens, dots, underscores, etc.) with a single `-`, then trim leading/trailing `-`". The worked example now traces correctly from the stated rule.
2. **Suffix part**: explicitly "literal `--rerec-<utc-timestamp>` where the trailing `Z` is uppercase, in UTC, and is part of the literal — do NOT lowercase or slug it".
3. **Function-call placeholder**: the inline placeholder is now `<slug-part>--rerec-<utc-timestamp-part>` with a "do NOT post-process the result" instruction, and the slug rule is fully declarative (no `slug()` token left to look up).

Also spelled out `<short-git-sha>` = first 7 chars of `git rev-parse HEAD` in the deterministic alternative, and noted that the alternative drops the sort-grouping property (so callers know what they're trading off).

versioned_docs/version-4.0.0/quickstart/k8s-proxy-llm-workflow.md
```diff
@@ -187,7 +188,64 @@ The contract changed on purpose; the test's recorded baseline is stale. Read `os
...
+   > 2. **Suffix part** — append the literal `--rerec-<utc-timestamp>`, where `<utc-timestamp>` is `YYYYMMDDHHMMZ` (the trailing `Z` is uppercase, in UTC, and is part of the literal — do NOT lowercase or slug it). Example suffix: `--rerec-202605281430Z`.
+   >
+   > Combined example: `scenario-4-v8-baseline-4-cases--rerec-202605281430Z`. Two agents running in different timezones at the same wall-clock minute produce the same name — preserving the "original + its re-records sort together" intent for cross-timezone CI integrations.
+   >
+   > **Deterministic alternative.** If the original-set-name isn't available at name-mint time, use `rerec-<short-git-sha>-<utc-timestamp>` where `<short-git-sha>` is the first 7 characters of `git rev-parse HEAD`. This form drops the original-set prefix, so re-records won't sort-group with the original on the recordings page — only use it when the prefix is genuinely unavailable.
```
charankamarapu: **[r4 — consistency — LOW] Separator inconsistency between primary form and deterministic alternative**

Primary form: `<slug-part>--rerec-<utc-timestamp-part>` — double dash before `rerec`, used as a slug/suffix boundary marker (so the recordings UI's prefix-match groups original + re-records together).

Deterministic alternative: `rerec-<short-git-sha>-<utc-timestamp>` — single dash before `rerec`, no leading prefix, no boundary marker.

The two forms answer different needs (with-prefix vs prefix-unavailable) but they diverge on the literal separator pattern around `rerec` for no semantic reason. An LLM agent reading both may get confused about which separator goes where:

- Agent A picks the alternative form correctly: `rerec-abc1234-202605281430Z`.
- Agent B reads the alternative but reflexively copies the primary's `--rerec-` literal (because that was the "canonical" pattern the doc led with): `--rerec-abc1234-202605281430Z`. The leading `--` looks like a CLI flag to log scrapers and shell-escaping pipelines — high chance of producing a bad shell command.

**Fix options:**
1. **Unify on `--rerec-` for both forms.** The alternative becomes `--rerec-<short-git-sha>-<utc-timestamp>` (with a noted-as-empty leading prefix). The literal becomes consistent across the two forms.
2. **Spell out the alternative's lack of leading separator explicitly.** Add one line: "Note: the deterministic alternative has no leading prefix or `--` — it begins with the literal `rerec`."

Option 2 is the smaller change. Either makes the two forms unambiguously distinguishable for an LLM agent following the spec mechanically.

charankamarapu: Fixed in `487144f2f09379570ee315ff9cac8726b29fc770`. Went with option 2 (lighter change). Added one parenthetical to the deterministic-alternative description that explicitly notes: "The alternative begins with the literal `rerec` (NO leading `-` or `--` — the double-dash in the primary form is the slug/suffix boundary marker, which this form doesn't have because there's no slug part)."

This makes the asymmetry intentional and explained, so an agent who reads the alternative knows the primary's `--rerec-` doesn't carry over. Keeping the two forms structurally distinct also signals their different intents (with-prefix vs prefix-unavailable) more clearly than a forced `--rerec-` unification would.

gouravkrosx (approved):

PR #880, opened 2026-06-26
update: CI_CD Docs

pathakharshit:
## Summary

Adds a **Running cloud replay in CI** section to all three CI/CD docs (GitHub, GitLab, Jenkins).

- Cloud replay re-runs **test sets recorded from a Kubernetes deployment** — this is now explicitly called out in each file
- Works with both Keploy Cloud and self-hosted Keploy
- Generic across CI systems: GitHub Actions, GitLab CI, and Jenkins examples included
- Each file has a system-specific, copy-pasteable pipeline snippet

## Changes

- `github.md` — new section with complete GitHub Actions workflow (name, on, jobs) and secrets.KEPLOY_API_KEY
- `gitlab.md` — new section with GitLab CI masked variable example, shell continuation with `|` and `\`
- `jenkins.md` — new section with Jenkins withCredentials example

## Test plan

- [x] Prettier passes on all three files
- [x] Vale spell check passes
- [x] commitlint passes (all commits use valid Conventional Commits types)
- [x] DCO signed off
- [x] Addressed all review comments from amaan-bhati: secrets context fixed, duplicate section removed, full YAML added, GitLab multi-line command style fixed, k8s-recorded tests clarification added

### DIFF
--- a/versioned_docs/version-4.0.0/ci-cd/github.md
+++ b/versioned_docs/version-4.0.0/ci-cd/github.md
@@ -280,3 +280,53 @@ keploy cloud replay --app "<NAMESPACE>.<DEPLOYMENT>" --cluster "<CLUSTER>" --nam
 > `--delay` sets how long Keploy waits for the application to become ready before it sends requests. If it is shorter than the application's startup time, the tests can fail, so set it to comfortably cover the boot time.
 
 Hope this helps you out, if you still have any questions, reach out to us .
+
+---
+
+## Running cloud replay in CI
+
+Keploy cloud replay re-runs your recorded test sets from a CI pipeline. It works with both **Keploy Cloud** and a **self-hosted Keploy** setup — the command is the same either way.
+
+### How authentication works
+
+The CLI reads the `KEPLOY_API_KEY` environment variable automatically. You do not need to pass it as a flag or log in through a browser.
+
+- Locally: `export KEPLOY_API_KEY="<your-api-key>"` before running the command.
+- In CI: add the key as a secret in your CI settings so the system injects it as an environment variable at runtime. Never hardcode it in your pipeline file.
+
+In GitHub Actions, secrets are stored under **Settings → Secrets and variables → Actions** and referenced in the pipeline as `${{ secrets.KEPLOY_API_KEY }}`.
+
+### Steps
+
+1. Add `KEPLOY_API_KEY` as a repository secret in GitHub (**Settings → Secrets and variables → Actions**).
+2. Install the Enterprise Keploy binary on the runner.
+3. Run `keploy cloud replay` with your application and cluster details.
+
+> Cloud replay requires the Enterprise binary (`keploy.io/ent/dl/latest/enterprise_linux_amd64`), not the open-source one.
+
+### Example: GitHub Actions
+
+```yaml
+jobs:
+  keploy-cloud-replay:
+    runs-on: ubuntu-latest
+    env:
+      KEPLOY_API_KEY: ${{ secrets.KEPLOY_API_KEY }}
+    steps:
+      - name: Install Keploy Enterprise
+        run: |
+          curl --silent --location "https://keploy.io/ent/dl/latest/enterprise_linux_amd64" -o /tmp/keploy
+          sudo chmod +x /tmp/keploy && sudo mv /tmp/keploy /usr/local/bin/keploy
+
+      - name: Cloud replay
+        run: |
+          keploy cloud replay \
+            --app "<NAMESPACE>.<DEPLOYMENT>" \
+            --cluster "<CLUSTER>" \
+            --namespace "<NAMESPACE>" \
+            --delay <DELAY>
+```
+
+Replace `<NAMESPACE>`, `<DEPLOYMENT>`, `<CLUSTER>`, and `<DELAY>` with your own values. Set `<DELAY>` to cover your application's startup time (in seconds).
+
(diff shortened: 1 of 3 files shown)


versioned_docs/version-4.0.0/ci-cd/github.md
```diff
@@ -215,4 +217,66 @@ sudo -E keploy test -c node src/app.js --delay 10 --path ./
...
+  --delay <DELAY>
+```
+
+Replace `<NAMESPACE>`, `<DEPLOYMENT>`, and `<CLUSTER>` with your own values, and set `<DELAY>` to cover your application's startup time.
+
+### Example: GitHub Actions
+
+```yaml
+jobs:
+  keploy-cloud-replay:
```
amaan-bhati, line 254: This snippet starts with `jobs:` so it's incomplete as a standalone workflow file. Someone creating a new `.yml` from this won't have a `name:` or `on:` trigger - the workflow will never actually run. Either note that this is meant to be added to an existing workflow, or show a more complete example with at least a `push` trigger.

versioned_docs/version-4.0.0/ci-cd/github.md
```diff
@@ -215,4 +217,66 @@ sudo -E keploy test -c node src/app.js --delay 10 --path ./
...
+            --delay <DELAY>
+```
+
+### Other CI systems
+
+The steps are identical elsewhere — expose `KEPLOY_API_KEY` from your secret store, install the binary, then run the command. For example, GitLab CI uses a masked CI/CD variable and Jenkins uses a credentials binding:
+
+```bash
+export KEPLOY_API_KEY="$KEPLOY_API_KEY" # injected from your CI secret store
+curl --silent --location "https://keploy.io/ent/dl/latest/enterprise_linux_amd64" -o /tmp/keploy
```
amaan-bhati: This is a no-op. GitLab CI and Jenkins inject secrets into the shell environment directly, so `$KEPLOY_API_KEY` is already set when your script runs. `export VAR="$VAR"` just re-exports the same value and adds confusion. Drop this line - the snippet works without it.

versioned_docs/version-4.0.0/ci-cd/github.md
```diff
@@ -215,4 +217,66 @@ sudo -E keploy test -c node src/app.js --delay 10 --path ./
...
+sudo chmod +x /tmp/keploy && sudo mv /tmp/keploy /usr/local/bin/keploy
+keploy cloud replay --app "<NAMESPACE>.<DEPLOYMENT>" --cluster "<CLUSTER>" --namespace "<NAMESPACE>" --delay <DELAY>
+```
+
+> `--delay` sets how long Keploy waits for the application to become ready before it sends requests. If it is shorter than the application's startup time, the tests can fail, so set it to comfortably cover the boot time.
+
```
amaan-bhati, line 276: This note is useful but it's at the bottom, after the Other CI section. `--delay` first shows up in the generic command block much earlier. Move this right after the `Replace <NAMESPACE>...` line so readers see the explanation before they run the command.

amaan-bhati (commented):
Hey @pathakharshit Thanks for raising this pr. A few things to fix before this can merge:

**Blocking**

The env variable reference in the YAML is wrong:

```yaml
env:
  KEPLOY_API_KEY: ${{ KEPLOY_API_KEY }}
```

In GitHub Actions, secrets are accessed via the `secrets` context. It should be:

```yaml
env:
  KEPLOY_API_KEY: ${{ secrets.KEPLOY_API_KEY }}
```

As written, this would silently pass an empty value and the workflow would fail with no clear error.

---

**Structure issue**

The section jumps straight to `### 2. Add the workflow` with no `### 1.`. The API key setup block (`export KEPLOY_API_KEY="<API_KEY>"`) is floating above it without a heading.

Step 1 should walk the user through storing the key as a GitHub secret (repo Settings > Secrets and variables > Actions > New repository secret, name it `KEPLOY_API_KEY`). The `export` command is for local shell use, not CI - either remove it or clarify that distinction.

---

**Incomplete workflow snippet**

The text says "Create `.github/workflows/keploy-cloud-replay.yml`" which implies a full file, but the YAML block only contains `jobs:`. It is missing `name:` and `on:` triggers. Anyone copy-pasting this will get a broken workflow file. At minimum add:

```yaml
name: Keploy Cloud Replay

on:
  push:
    branches: [main]

jobs:
  keploy-cloud-replay:
    ...
```

---

**PR description**

Please fill out the PR description. It still has the template placeholder text with none of the checkboxes checked. A one-liner on what this adds would work.

---

Happy to re-review once these are addressed.

amaan-bhati (commented):
Thanks for the updates @pathakharshit One thing to fix in the GitLab snippet before this can merge.

In `gitlab.md`, the `keploy cloud replay` command is split across multiple lines inside `script:` without shell `\` continuation:

```yaml
    - keploy cloud replay
      --app "<NAMESPACE>.<DEPLOYMENT>"
      --cluster "<CLUSTER>"
      --namespace "<NAMESPACE>"
      --delay <DELAY>
```

This relies on YAML plain scalar folding to join the lines before passing them to the shell, which technically works but is non-idiomatic. Anyone reading it will assume these are separate shell commands, and anyone editing it might add a `\` thinking they need it - which would actually break it. The Jenkins example in this same PR does it correctly with `\`. Use the same approach here:

```yaml
    - |
      keploy cloud replay \
        --app "<NAMESPACE>.<DEPLOYMENT>" \
        --cluster "<CLUSTER>" \
        --namespace "<NAMESPACE>" \
        --delay <DELAY>
```

That aside, the rest looks good. Once addressed, please re request a review!

amaan-bhati (approved):
LGTM
Checked locally, ui looks fine:

Image

PR #881, opened 2026-06-26
docs(ci-cd): use enterprise install script for cloud replay

pathakharshit:
Replace the manual binary download (curl+chmod+mv) with the enterprise install script across all three CI/CD docs (github.md, gitlab.md, jenkins.md). Also updates the callout note in each file to reference the install script URL.


Aditya-eddy (approved):
LGTM

amaan-bhati:
Hey @pathakharshit there is one thing that i wanted to point out but the pr got merged before i could do so, hence can you kindly resolve this in a different pr:

The install command works fine in GitHub Actions and GitLab CI since both default to bash, but the Jenkins example will break.

`sh '''...'''` in a Declarative Pipeline runs `/bin/sh`, which is `dash` on Ubuntu - and `dash` does not have `source`. Running this as-is on Jenkins would give you `source: not found`.

Fix: replace `source` with `.` (the POSIX equivalent), which works in both sh and bash:

```sh
curl --silent -O -L https://keploy.io/ent/install.sh && . install.sh
```

This is the only thing to fix - the GitHub Actions and GitLab CI snippets are fine.

PR #882, opened 2026-06-26
fix(ci-cd): use POSIX dot operator instead of source in Jenkins snippet

pathakharshit:
The Jenkins sh block runs under /bin/sh (dash on Ubuntu), which does not have source. This causes source install.sh to fail with source: not found.

Fix: replace source with . (the POSIX dot operator), which is equivalent and works in both sh and bash.

Only jenkins.md is affected — GitHub Actions and GitLab CI both default to bash where source works fine.


amaan-bhati (approved):
LGTM

PR #245, opened 2023-10-22
docs: add pull flag to macos installation

0xVikasRushi:
# Pull Request Template

## Description

changed macos alias and added --pull always

Closes: [#1037](https://github.com/keploy/keploy/issues/1037)

## Type of change
To ensure that macOS users consistently obtain the latest version of Keploy, by adding the --pull always flag to the alias in the documentation.

PR #890, opened 2026-07-13
feat: add Meta Pixel to docs site

dhananjay6561:
## What

Installs the Meta (Facebook) Pixel (`2006330080011702`) site-wide on the docs.

- **Base snippet** injected into `<head>` on every page via `headTags` (`docusaurus.config.js`): loads `fbevents.js`, runs `fbq('init', ...)`, and fires the initial `PageView`. `fbq` is defined synchronously by the snippet, so the first `PageView` can never race the loader.
- **SPA route tracking** (`src/metaPixelRouteTracker.js`, registered as a `clientModule`): re-fires `PageView` on client-side navigation via Docusaurus's `onRouteDidUpdate` lifecycle hook. Docusaurus is a single-page app, so without this the base snippet would only ever report the first page a visitor lands on.
- Includes the `<noscript>` tracking-pixel fallback for JS-disabled clients.

## Events tracked

- **`PageView`** only.
- No other events (`ViewContent`, `Lead`, `CompleteRegistration`, custom, etc.) are tracked in this PR. Those can be added later on specific CTAs/flows.

## When does the event fire?

The **first** PageView is fired synchronously by the base script (where `fbq` is guaranteed defined), and **subsequent** PageViews are fired by the client module on route changes. This split avoids both an fbq-not-ready race on the first load and double-counting the initial page.

| User action | What happens | PageView fired? |
| --- | --- | --- |
| Lands on the site / full reload | `<head>` snippet runs `fbq('init')` + `fbq('track','PageView')` | ✅ 1 (base script) |
| `onRouteDidUpdate` runs on first render | `previousLocation` is `null` → guard skips it | ⏭️ skipped (avoids double-count) |
| Clicks an internal link (SPA nav) | `location.pathname` differs from `previousLocation.pathname` → fires `fbq('track','PageView')` | ✅ 1 per navigation |
| Navigation where only the query/hash changes (same path) | pathname unchanged → guard skips | ❌ none |
| Ad blocker / tracking protection on, or JS disabled | `fbevents.js` blocked (the `<noscript>` `` covers the JS-off case with a single beacon) | ⚠️ browser pixel dropped |

## How the pixel leads to an ad (flow)

```mermaid
flowchart TD
    A[User opens keploy.io/docs] --> B[Meta Pixel runs on page load<br/>fbq init + PageView]
    B --> C[Browser sends beacon to Meta<br/>facebook.com/tr + Facebook cookies]
    C --> D[Meta identifies the visitor via cookie<br/>Facebook account XYZ]
    D --> E[Meta adds the user to Keploy's audience]
    E --> F[Later the user scrolls Instagram / Facebook]
    F --> G[Meta serves the user a Keploy ad]
```

The site never sends personal data — it only says "a PageView happened." **Identity is matched on Meta's side via the Facebook cookies the browser attaches to the `/tr` request.** This depends on third-party cookies, so it's weaker/absent for logged-out users, Safari/Firefox, private mode, and ad blockers (a future Conversions API would recover some of that).

## How to test (for reviewers)

> **Important:** disable ad blockers / tracking protection first, or test in an Incognito window with the ad blocker set to *not* run in Incognito. Blockers drop `fbevents.js` (shows as `blocked:other`) and you'll see zero events even though the code is correct.

1. Install the [Meta Pixel Helper](https://chromewebstore.google.com/detail/meta-pixel-helper/fdgfkebogiimcoedlicjlajpkdmockpc) Chrome extension.
2. Run the docs locally: `yarn start`, then open `http://localhost:3000/docs/`.
3. Open DevTools → **Network** tab, check **"Preserve log"**, and type `facebook` in the filter bar.
4. **Reload the page.** Expected:
   - `fbevents.js` → `200` (not `blocked:other` / 0 kB)
   - `tr/?id=2006330080011702&ev=PageView` → `200`
   - Meta Pixel Helper shows the pixel **Active** with **1 PageView**.
5. **Click an internal link** (SPA navigation, no full reload). Expected:
   - a **new** `tr/?...ev=PageView` beacon appears → PageView count goes **1 → 2**.
6. **Click a link to a same-path URL (query/hash only).** Expected: **no** new beacon (pathname-guarded).
7. *(Optional, end-to-end)* In **Meta Events Manager → Test Events**, enter pixel `2006330080011702` and browse the local site to watch the same PageViews land on Meta's side.

**Pass criteria:** 1 PageView on load, +1 per SPA path change, 0 on same-path clicks — all beacons `200`.

## Verify (author)

- Ran a full `yarn build` locally; compiles clean and the generated static HTML contains the pixel (`fbq("init","2006330080011702"),fbq("track","PageView")`, the `connect.facebook.net/en_US/fbevents.js` loader, and the `<noscript>` beacon).
- `onRouteDidUpdate` guard reviewed: fires only on an actual `pathname` change, skips the initial load (`previousLocation === null`) and same-path updates — so no double-count on landing and one PageView per SPA navigation.
- No CSP changes needed for docs (no enforcing Content-Security-Policy blocks Facebook's domains here); this PR touches only `docusaurus.config.js` and `src/metaPixelRouteTracker.js`.

## Note on the hardcoded pixel ID

The pixel ID is inlined, matching every other tracker in this repo — Microsoft Clarity (`static/scripts/clarity.js`), Apollo (`apollo-init.js`), and Google Analytics/gtag (`docusaurus.config.js`) are all hardcoded, and the repo has no `.env` files or `process.env` usage to read from. The preview/staging-pollution concern is valid but applies equally to those existing trackers, so moving all of them behind env config (with preview/staging skipped) is proposed as a dedicated repo-wide follow-up rather than gating the pixel alone here.

Docs deploys separately from the landing page ([keploy/landing#245](https://github.com/keploy/landing/pull/245) adds the same snippet there) — this PR gives the docs site the equivalent coverage.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -568,6 +568,7 @@ fbq('track', 'PageView');`,
     ],
   ],
 
+  clientModules: [require.resolve("./src/metaPixelRouteTracker.js")],
   scripts: [
     {
       src: "/docs/scripts/feedback.js",
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -173,7 +173,7 @@ fbq('track', 'PageView');`,
     {
       tagName: "noscript",
       attributes: {},
-      innerHTML: `<img height="1" width="1" style="display:none" src="https://www.facebook.com/tr?id=2006330080011702&ev=PageView&noscript=1" />`,
+      innerHTML: `<img height="1" width="1" style="display:none" alt="" src="https://www.facebook.com/tr?id=2006330080011702&ev=PageView&noscript=1" />`,
     },
     // End Meta Pixel Code
   ],
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -155,6 +155,25 @@ module.exports = {
         },
       }),
     },
+    // Meta Pixel Code
+    {
+      tagName: "script",
+      innerHTML: `!function(f,b,e,v,n,t,s)
+{if(f.fbq)return;n=f.fbq=function(){n.callMethod?
+n.callMethod.apply(n,arguments):n.queue.push(arguments)};
+if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0';
+n.queue=[];t=b.createElement(e);t.async=!0;
+t.src=v;s=b.getElementsByTagName(e)[0];
+s.parentNode.insertBefore(t,s)}(window, document,'script',
+'https://connect.facebook.net/en_US/fbevents.js');
+fbq('init', '2006330080011702');
+fbq('track', 'PageView');`,
+    },
+    {
+      tagName: "noscript",
+      innerHTML: `<img height="1" width="1" style="display:none" src="https://www.facebook.com/tr?id=2006330080011702&ev=PageView&noscript=1" />`,
+    },
+    // End Meta Pixel Code
   ],
   title: "Keploy Documentation",
   titleDelimiter: "🐰",
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -158,6 +158,7 @@ module.exports = {
     // Meta Pixel Code
     {
       tagName: "script",
+      attributes: {},
       innerHTML: `!function(f,b,e,v,n,t,s)
 {if(f.fbq)return;n=f.fbq=function(){n.callMethod?
 n.callMethod.apply(n,arguments):n.queue.push(arguments)};
@@ -171,6 +172,7 @@ fbq('track', 'PageView');`,
     },
     {
       tagName: "noscript",
+      attributes: {},
       innerHTML: `<img height="1" width="1" style="display:none" src="https://www.facebook.com/tr?id=2006330080011702&ev=PageView&noscript=1" />`,
     },
     // End Meta Pixel Code
--- a/src/metaPixelRouteTracker.js
+++ b/src/metaPixelRouteTracker.js
@@ -4,7 +4,7 @@
 // explicitly here.
 export function onRouteDidUpdate({location, previousLocation}) {
   if (previousLocation && location.pathname !== previousLocation.pathname) {
-    if (typeof window !== "undefined" && typeof window.fbq === "function") {
+    if (typeof window.fbq === "function") {
       window.fbq("track", "PageView");
     }
   }
--- a/src/metaPixelRouteTracker.js
+++ b/src/metaPixelRouteTracker.js
@@ -1,0 +1,11 @@
+// Re-fire Meta Pixel PageView on client-side route changes.
+// The base snippet in docusaurus.config.js (headTags) only fires PageView on
+// the initial load; Docusaurus is an SPA, so in-app navigation must be tracked
+// explicitly here.
+export function onRouteDidUpdate({location, previousLocation}) {
+  if (previousLocation && location.pathname !== previousLocation.pathname) {
+    if (typeof window !== "undefined" && typeof window.fbq === "function") {
+      window.fbq("track", "PageView");
+    }
+  }
+}


amaan-bhati (commented):
**Build not verified locally**

The PR description calls out that a full Docusaurus build was not run due to disk space. Worth someone on the team doing a quick npm run build on this branch before merging, or at minimum checking the Network tab for the tr?...ev=PageView beacon after the preview deploy lands.

docusaurus.config.js
```diff
@@ -155,6 +155,27 @@ module.exports = {
...
+      attributes: {},
+      innerHTML: `!function(f,b,e,v,n,t,s)
+{if(f.fbq)return;n=f.fbq=function(){n.callMethod?
+n.callMethod.apply(n,arguments):n.queue.push(arguments)};
+if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0';
+n.queue=[];t=b.createElement(e);t.async=!0;
+t.src=v;s=b.getElementsByTagName(e)[0];
+s.parentNode.insertBefore(t,s)}(window, document,'script',
+'https://connect.facebook.net/en_US/fbevents.js');
+fbq('init', '2006330080011702');
```
amaan-bhati, line 170: **Hardcoded pixel ID**

Same concern as the landing site PR. Every Vercel preview deployment fires real events into the production pixel. Worth reading from an environment variable and skipping the init entirely when it is not set, so staging and preview traffic stays out of the production audience.

docusaurus.config.js
```diff
@@ -155,6 +155,27 @@ module.exports = {
...
+fbq('track', 'PageView');`,
+    },
+    {
+      tagName: "noscript",
+      attributes: {},
+      innerHTML: `<img height="1" width="1" style="display:none" src="https://www.facebook.com/tr?id=2006330080011702&ev=PageView&noscript=1" />`,
```
amaan-bhati: **Missing alt on noscript image**

The landing page implementation has alt="" on this img. Standard practice for a tracking pixel even with display:none. Worth adding for consistency.

src/metaPixelRouteTracker.js
```diff
@@ -0,0 +1,11 @@
+// Re-fire Meta Pixel PageView on client-side route changes.
+// The base snippet in docusaurus.config.js (headTags) only fires PageView on
+// the initial load; Docusaurus is an SPA, so in-app navigation must be tracked
+// explicitly here.
+export function onRouteDidUpdate({location, previousLocation}) {
+  if (previousLocation && location.pathname !== previousLocation.pathname) {
+    if (typeof window !== "undefined" && typeof window.fbq === "function") {
```
amaan-bhati: **Redundant window check**

Docusaurus client modules only run in the browser, so typeof window !== "undefined" is always true here. The check that matters is typeof window.fbq === "function".

docusaurus.config.js
dhananjay6561: Fixed in 77f5c130 — added `alt=""` to the `<noscript>` tracking pixel to match the landing implementation. Confirmed it renders in the production build.

src/metaPixelRouteTracker.js
dhananjay6561: Fixed in 77f5c130 — dropped the `typeof window !== "undefined"` guard and kept only `typeof window.fbq === "function"`. You're right, client modules only ever run in the browser so that check was dead weight.

docusaurus.config.js
dhananjay6561: Good call, and agreed it is the cleaner setup. Holding off on env-gating in this PR for two reasons:

1. **Consistency with landing** — the landing PR (#245) also hardcodes the same ID inline, so gating it only here would leave the two repos diverging on the same pixel.
2. **Merge-time risk** — if I switch to `process.env.META_PIXEL_ID` and skip init when unset, the pixel silently goes dark in production the moment this merges *unless* the env var is already wired in Vercel for docs. So this needs the Vercel env set up first, ideally across both repos together.

Proposing a follow-up to standardize env-based config (with preview/staging skipped) on landing + docs in one go. Happy to do it in this PR instead if you want it here now — just say the word and confirm the Vercel `META_PIXEL_ID` var is set for the docs project.

dhananjay6561: Dug into how the repo handles its other trackers, and hardcoding is actually the established pattern here — env-gating the pixel alone would be the inconsistent one:

| Tool | ID | Where | Env var? |
| --- | --- | --- | --- |
| Microsoft Clarity | `8tes8khpwo` | `static/scripts/clarity.js` | ❌ hardcoded |
| Apollo tracker | `6644a0d6a54b5b0438c841cc` | `apollo-init.js` | ❌ hardcoded |
| Google Analytics (gtag) | `G-LLS95VWZPC` | `docusaurus.config.js` | ❌ hardcoded |
| Meta Pixel (this PR) | `2006330080011702` | `docusaurus.config.js` | ❌ hardcoded |

`process.env` is used 0 times in `docusaurus.config.js` and there are no `.env` files in the repo — so there is no env-config plumbing to read from yet.

The preview/staging-pollution concern is completely valid, but it applies equally to GA, Clarity, and Apollo, which all fire on every preview deploy today. So it is really a repo-wide (and landing-wide, since #245 hardcodes too) decision, not something to solve for the pixel alone in this PR.

Proposing a dedicated follow-up to move all four trackers behind env config with preview/staging skipped, so it is done consistently in one place. Happy to own that. If you would rather gate just the pixel here now, I can — just confirm the Vercel `META_PIXEL_ID` var is set for the docs project first, otherwise the pixel goes dark in prod on merge.

amaan-bhati (approved):
LGTM

dhananjay6561:
cc: @nehagup

PR #891, opened 2026-07-17
feat: add llms markdown in doc pages

shreya1510ss:
## What has changed?

Every docs page is now also available as plain Markdown, and `llms.txt` / `llms-full.txt` are auto-generated at build time — so AI agents (and humans) can grab clean per-page content instead of parsing HTML.

- Added `docusaurus-plugin-llms`, configured to generate `llms.txt`, `llms-full.txt`, and a `.md` twin for every doc page (same URL, `.md` extension).
- Added a "Copy as Markdown" / "View as Markdown" control under each page title, so the `.md` file is actually discoverable, not just guessable. Styled with Tailwind utility classes to match this repo's existing convention.
- Patched `docusaurus-plugin-llms` (via `patch-package`, auto-reapplied on install) to fix two real content bugs found during review:
  - Some pages reuse content from another file via MDX imports (e.g. `import WhatIsKeploy from '../concepts/what-is-keploy.md'`); the plugin only resolved these when the filename had an underscore, so ~7 pages were leaking that raw import/JSX line into their `.md` output instead of the real text.
  - ~106 pages were also leaking raw React component code (`<Tabs>`, `<ProductTier>`, `<HowTo>`, etc.) into their `.md` output. Fixed by stripping self-closing component tags entirely and unwrapping paired ones while preserving their real inner content — carefully excluding `<YOUR_INGRESS_HOST>`-style fill-in-your-own-value placeholders, which look similar but aren't components.
- Fixed a timer leak in the copy button: a `setTimeout` resetting its label was never cleared, so it could fire after the component unmounted (React warning) or race itself on repeated clicks.

Fixes https://github.com/keploy/enterprise/issues/2272

## How Has This Been Tested?

Ran `npm run build` then `npm run serve` — confirmed the build succeeds, `llms.txt`/`llms-full.txt` generate correctly, each page's `.md` twin loads at its matching URL, and the Copy/View buttons work in both light and dark mode. 

https://github.com/user-attachments/assets/c9408f3b-1e19-4ef0-8e1d-415a15c45c89


src/components/MarkdownPageActions.js
```diff
@@ -0,0 +1,76 @@
...
+      if (!res.ok) throw new Error(`Request failed: ${res.status}`);
+      const text = await res.text();
+      await navigator.clipboard.writeText(text);
+      setCopyState("copied");
+    } catch (e) {
+      setCopyState("error");
+    } finally {
+      setTimeout(() => setCopyState("idle"), 2000);
+    }
+  };
```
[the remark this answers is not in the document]
shreya1510ss: resolved this.

src/theme/DocItem/index.js
```diff
@@ -171,6 +172,13 @@ export default function DocItem(props) {
   const isCategoryIndex =
     frontMatter?.slug === "index" || /\/category\/|\/index\/?$/.test(permalink);
   const suppressArticleSchema = isDocsRoot || isCategoryIndex;
+  // .md twins are only generated for the current (4.0.0) version's docs
+  // (docusaurus-plugin-llms docsDir points at versioned_docs/version-4.0.0),
+  // and only for real doc pages, not root/category-index pages.
+  const showMarkdownActions = isLatestVersion && !suppressArticleSchema;
+  const markdownUrl = showMarkdownActions
+    ? `${permalink.replace(/\/+$/, "")}.md`
```
manasmanohar: Copy/View as Markdown 404s on the 5 section landing pages (concepts, dependencies, operation, server, server/sdk-installation). For an index page the permalink is `/docs/concepts/`, so this produces `/docs/concepts.md` — but the plugin writes the file to `/docs/concepts/index.md`, so the link 404s and Copy shows "Copy failed". Normal pages are fine.
Options: special-case folder permalinks (trailing slash → append `index.md` instead of `.md`), or derive the .md URL from the plugin's output instead of reconstructing it from the permalink.

patches/docusaurus-plugin-llms+0.4.0.patch
```diff
@@ -0,0 +1,68 @@
...
++    // self-closing pass must run first, or the second regex would partially match it.
++    //
++    // Docs also use <YOUR_INGRESS_HOST>-style ALL_CAPS placeholders for readers to
++    // fill in their own values — these also start with a capital letter, so a real
++    // component name is only treated as such when it contains a lowercase letter
++    // and no underscore (true of every actual component in this codebase, never
++    // true of these placeholders).
++    const isRealComponentName = (name) => /[a-z]/.test(name) && !name.includes('_');
++    // Prop values can legitimately contain '<' or '>' inside quoted strings (e.g.
++    // tools={["Linux kernel >= 5.10"]}), which would otherwise be mistaken for the
```
manasmanohar, line 62: These two regexes run over the whole document including fenced code, so real code gets stripped. e.g. `installation/windows.md`: `wsl --install -d <Distribution Name>` → `wsl --install -d` (~5 files); `quickstart/java-spring-boot-xml.md`: the `<User>...</User>` wrapper is removed and the XML sample ends up malformed. This is new — those tags were preserved before (the plugin only stripped a fixed HTML allowlist).
Option: skip fenced code blocks when stripping (mask fences → strip prose → restore).

package.json
```diff
@@ -58,14 +59,17 @@
     "@tailwindcss/typography": "^0.5.0",
     "autoprefixer": "^10.4.0",
     "docusaurus": "^1.14.7",
+    "docusaurus-plugin-llms": "^0.4.0",
```
manasmanohar: Plugin is `^0.4.0` but the patch is pinned to `0.4.0` — patch-package will fail the install if a `0.4.1` publishes. Consider pinning the exact version.

docusaurus.config.js
```diff
@@ -178,6 +178,18 @@ module.exports = {
       };
     },
     "docusaurus-tailwindcss-loader",
+    [
+      "docusaurus-plugin-llms",
+      {
+        docsDir: "versioned_docs/version-4.0.0",
```
manasmanohar: `docsDir` is hardcoded to `versioned_docs/version-4.0.0` — will need updating when we cut the next docs version, since the buttons show on whatever the latest version is (gated on `isLatestVersion`).

manasmanohar (commented):
Please address the comments below. Specifics left inline; this one isn't tied to a diff line:

**Collides with the llms.txt we already ship.**
We already have curated `static/llms.txt` and `static/llms-full.txt` (content team's — About Keploy, awards, positioning, curated links). The plugin generates the same two filenames, and its postBuild runs after `static/` is copied, so it overwrites them: `build/llms.txt` is the generated version and the curated content is gone. Net effect — the content team's file is silently replaced, and editing `static/llms.txt` later would have no effect. The static file also drifts today since it's hand-maintained.
Option: merge both — feed the curated text into the plugin via `rootContent`/`fullRootContent` (+ `title`/`description`) so it stays on top and the auto-generated index below stays fresh. Needs the curated source moved out of `static/` (so it isn't overwritten) and the static copies removed. Worth confirming the exact split with the content team.

src/components/MarkdownPageActions.js
```diff
@@ -0,0 +1,82 @@
...
+
+  const copyLabel =
+    copyState === "copying"
+      ? "Copying…"
+      : copyState === "copied"
+        ? "Copied"
+        : copyState === "error"
+          ? "Copy failed"
+          : "Copy as Markdown";
```
manasmanohar, line 55: add fixed width to prevent the div from shrinking when changing text

src/theme/DocItem/index.js
shreya1510ss: this has been resolved. added both the conditions for generating the md link.

src/components/MarkdownPageActions.js
shreya1510ss: fixed the width .

package.json
shreya1510ss: pinned the exact version.

patches/docusaurus-plugin-llms+0.4.0.patch
shreya1510ss: done.

docusaurus.config.js
```diff
@@ -199,6 +199,18 @@ fbq('track', 'PageView');`,
...
+        ignoreFiles: ["**/shared/**"],
+        generateLLMsTxt: true,
+        generateLLMsFullTxt: true,
+        generateMarkdownFiles: true,
+        preserveDirectoryStructure: false,
+        excludeImports: true,
```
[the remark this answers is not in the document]
shreya1510ss: docsDir is intentionally pointed at versioned_docs/version-4.0.0 rather than docs/ — this repo uses full Docusaurus versioning with lastVersion: "4.0.0" set in docusaurus.config.js. Under that setup, the bare /docs/... URLs are served from whichever folder is lastVersion, not from docs/. In fact docs/ here only contains a few non-versioned pages (components, gsoc, hacktoberfest) — none of the real content (concepts, quickstart, installation, etc.) lives there; it's all under versioned_docs/version-4.0.0/.

nehagup:
Can we please add canonical tags so that these markdown pages doesn't cannibalise with github docs md links?

shreya1510ss:
@nehagup Looked into this — a few points:

1. Canonical <link> tags can't be applied to the .md files themselves (plain text, no <head>). Our HTML pages already have canonical tags via Docusaurus.
2. Checked how Stripe and Sarvam handle the same setup — both use an X-Robots-Tag: noindex header on the .md response instead of a canonical tag. Keeps it readable for AI tools, excludes it from Google indexing.
3. We're served via CloudFront (not Vercel), so the fix is a CloudFront Response Headers Policy adding X-Robots-Tag: noindex on /docs/*.md.

manasmanohar:
> @nehagup Looked into this — a few points:
> 
> 1. Canonical  tags can't be applied to the .md files themselves (plain text, no ). Our HTML pages already have canonical tags via Docusaurus.
> 2. Checked how Stripe and Sarvam handle the same setup — both use an X-Robots-Tag: noindex header on the .md response instead of a canonical tag. Keeps it readable for AI tools, excludes it from Google indexing.
> 3. We're served via CloudFront (not Vercel), so the fix is a CloudFront Response Headers Policy adding X-Robots-Tag: noindex on /docs/*.md.

sounds good, lets make the changes required in cloudfront.

src/components/MarkdownPageActions.js
```diff
@@ -0,0 +1,82 @@
...
+  if (!mdUrl) return null;
+
+  const handleCopy = async () => {
+    clearTimeout(resetTimeoutRef.current);
+    setCopyState("copying");
+    try {
+      const res = await fetch(mdUrl);
+      if (!res.ok) throw new Error(`Request failed: ${res.status}`);
+      const text = await res.text();
+      await navigator.clipboard.writeText(text);
```
manasmanohar: Copy is failing on safari because WebKit only allows navigator.clipboard.writeText() while the user activation from the click is still active. 

https://github.com/user-attachments/assets/e0491677-4f57-448d-9ea7-4045e1a7d97a

Our current implementation performs the clipboard write after an async boundary, so Safari rejects it with NotAllowedError. Chrome and Firefox are more permissive, which is why the issue wasn’t visible there.

manasmanohar (changes requested):
please address the comments, rest looks good!

src/components/MarkdownPageActions.js
manasmanohar: add logs to the catch the error  
```[MarkdownPageActions] copy failed:
NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
writeText
(anonymous function) — 17896441.5827e3b9.js:192```

manasmanohar (approved):
LGTM!

PR #895, opened 2026-08-03
docs: update download URLs from S3 to Azure

slayerjain:
Download hosting has moved from the keploy-enterprise S3 bucket (us-west-2) to Azure: user-facing download links now use keploy.io/ent/dl and Dockerfile ADD snippets and agent JARs use the keployenterprise blob storage directly. The S3 bucket is being decommissioned. All rewritten URLs were verified serving (HTTP 206).

### DIFF
--- a/versioned_docs/version-2.0.0/keploy-cloud/time-freezing.md
+++ b/versioned_docs/version-2.0.0/keploy-cloud/time-freezing.md
@@ -56,7 +56,7 @@ uname -a
 
 ```Dockerfile
 # Download the time freeze agent
-ADD https://keploy-enterprise.s3.us-west-2.amazonaws.com/releases/latest/assets/go_freeze_time_amd64 /lib/keploy/go_freeze_time_amd64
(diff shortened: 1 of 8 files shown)


slayerjain:
Independent review pass: no blocking findings — correctness, URL encoding/serving, and consumer semantics verified.

Minor notes (non-blocking):
- `versioned_docs/version-4.0.0/running-keploy/keploy-karaf.md`: Six pages hard-code the provider hostname keployenterprise.blob.core.windows.net, recreating the exact vendor-URL churn this PR exists to fix; the keploy.io/ent/dl redirector already covers the releases container (so the time-freezing Dockerfile ADDs could use provider-agnostic URLs today), and the

slayerjain:
Heads-up on priority: the live time-freezing page this PR updates (https://keploy.io/docs/keploy-cloud/time-freezing/) is linked from shipped CLI diagnostics (enterprise `pkg/service/test/test.go`), and its current S3 ADD snippets will break once the S3 bucket is emptied. This needs to merge and the site redeploy before the S3 decommission — see keploy/enterprise#2342.

PR #896, opened 2026-08-07
perf(docs): core web vitals — self-host fonts, drop dead assets, defer pixel, add security headers

dhananjay6561:
## Summary

Core Web Vitals + web-quality fixes for the **docs** site, from the Aug 2026 web-quality action plan (docs tickets **D1–D11** and the docs-relevant cross-cutting **X/H** items).

Docs mobile was failing **LCP (~2.6s)** mainly on render-blocking web fonts and heavy GIFs. This PR lands every ticket that can be done **without a new runtime dependency and without breaking anything** — all build-verified. Genuinely-blocked tickets (and why) are listed at the bottom so nothing is silently dropped.

> **Scope:** docs repo only. All `landing` (L*) and `blog` (B*) tickets live in their own repos and are out of scope here.

---

## 📊 Ticket status

Every docs ticket from the action plan, verified on the served version (**v4.0.0** — `CURRENT_DOCS_VERSION`; v3.0.0 is not built). `File` verified on this branch.

| ID | Area | What | Key file(s) | Status |
|---|---|---|---|---|
| **D1** | `LCP` | Self-host DM Sans (variable woff2, `latin`+`latin-ext`); drop render-blocking Google Fonts `<link>` | `src/fonts/DMSans-*.woff2`, `src/css/custom.css`, `docusaurus.config.js` | ✅ Done |
| **D2** | `LCP` `DEAD` | Delete 6 zero-ref heavy assets (**−62 MB**) | `static/{gif,img,cms}` | ✅ Done |
| **D3** | `LCP` | Convert 4 in-use GIFs → H.264 MP4 (**−78%**); swap 30 refs across 24 files to `<video>` | `static/gif`, `static/img`, 24× `*.md` | ✅ Done |
| **D4** | `LCP` | Analytics loading refactor: **GA + Meta Pixel eager** (accuracy — see note), **Clarity + Apollo interaction-gated**, **Hotjar removed** | `src/metaPixelRouteTracker.js`, `docusaurus.config.js` | ✅ Done |
| **D5** | `CLS` | Stamp intrinsic `width`/`height` on raw `` (dependency-free remark plugin) | `src/remark/remarkImageSize.js`, `docusaurus.config.js` | ✅ Done |
| **D6** | `LCP` | Preconnects **6 → 4** (follows from D1) | `docusaurus.config.js` | ✅ Done |
| **D9** | `A11Y` | `aria-hidden` on 5 footer social + 12 decorative component SVGs + 6 inline SVGs in the GSoC guide | `src/theme`, `src/components`, `gsoc/contribution-guide.md` | ✅ Done |
| **D10** | `DEAD` | Remove undefined `"Aeonik"` font-family | `src/css/custom.css` | ✅ Done |
| **D11** | `SEC` | Client source maps (`devtool: source-map`, client-only) | `docusaurus.config.js` | ✅ Done |
| **X1** | `PROC` | CI guard — fail on new image > 500 KB | `.github/workflows/asset-budget.yml` | ✅ Done |
| **X3** | `PROC` | CI guard — fail on new `fonts.googleapis.com/css` link | `.github/workflows/asset-budget.yml` | ✅ Done |
| **X4** | `A11Y` | Accessibility-tree well-formedness for AI crawlers | resolved by D9 | ✅ Done |
| **H1** | `SEO` | robots.txt / sitemap / canonical / JSON-LD present & valid | — | ✅ Verified, no change |
| **H2** | `SEO` | Product / review schema | — | ⛔ Won't do — see below |

---

## 📊 Measured performance (Lighthouse, mobile, **median of 9 runs**)

Both rows were measured **back-to-back on the same machine**, so the delta is apples-to-apples. Lab scores are noisy (**±10–15 pts** run-to-run, and absolute values shift with machine/network/CDN state) — the **delta** is the reliable signal, not any single number.

| | Perf | LCP | TBT | CLS |
|---|---|---|---|---|
| **Production** (live `keploy.io/docs`) | 61 (53–65) | 5.1s | ~450ms | 0 |
| ✅ **This PR (shipped)** | **81** (70–84) | **3.1s** | ~478ms | 0 |

**Desktop (shipped):** **99** (96–100) / LCP **0.8s** / CLS 0.

**Net: +20 mobile perf, −2.0s LCP, CLS a perfect 0.** Best mobile runs hit 84 / LCP 2.9s; desktop is fully green. Mobile LCP (3.1s) is still above the 2.5s line — the residual is Docusaurus's own React hydration (771 KB) plus the two **eager** trackers (GA + Meta Pixel), kept eager by product requirement for analytics accuracy.

### ⚖️ Analytics loading strategy
- **GA + Meta Pixel** → **eager** (accurate analytics/conversion from first paint; product requirement).
- **Clarity + Apollo** → on **first user interaction** (engaged sessions only).
- **Hotjar** → **removed** (redundant session-recorder, ~56 KiB).
- **keploy telemetry** → eager (first-party, ~2 KiB).

> Idle-deferring GA + Pixel too would gain a few more mobile points, but they'd fire ~1–3s after paint — the team chose **instant** firing. Everything else (fonts, GIF→video, dead-asset removal, image dimensions, a11y, Hotjar removal, Clarity/Apollo gating) is applied.

### Performance — LCP

**D1 · Self-host DM Sans**
DM Sans loaded via a render-blocking `<link rel="stylesheet">` to `fonts.googleapis.com` in `headTags`. Now self-hosted as a variable woff2 (`latin` + `latin-ext` subsets) in `src/fonts/`, wired via `@font-face` in `custom.css` — mirroring the existing Roboto setup (`font-display: swap`). It's the same font Google serves modern browsers (DM Sans v17 variable), so letterforms are identical; only the source changes.

**D6 · Fewer preconnects**
Dropped the two now-unused font preconnects. Preconnects **6 → 4** (algolia, keploy.io, GA, GTM) — clears the ">4 preconnect" warning.

**D2 · Delete dead heavy assets (~62 MB)**
`git rm` of 6 assets confirmed **0-ref** (grepped repo-wide first):

| File | Size |
|---|---|
| `static/gif/unit-test.gif` | 24.8 MB |
| `static/img/unit-test.gif` | 24.8 MB |
| `static/img/record-testcase.gif` | 6.0 MB |
| `static/gif/interoperability.gif` | 2.4 MB |
| `static/gif/tc-generation.gif` | 1.4 MB |
| `static/cms/reactor.png` | 2.7 MB |

**D3 · Convert heavy in-use GIFs to H.264 MP4 (7.44 MB → 1.60 MB, −78%)**
The four *referenced* GIFs over 1 MB were the largest remaining LCP/bandwidth cost. Each was re-encoded to H.264 MP4 with `faststart` (recipe R3) and the GIF deleted:

| Asset | GIF | MP4 | Saved |
|---|---|---|---|
| `gif/record-replay` | 2.66 MB | 0.28 MB | −89% |
| `gif/replay-tc` | 2.19 MB | 0.41 MB | −81% |
| `gif/how-keploy-works` | 1.45 MB | 0.14 MB | −90% |
| `img/record-api` | 1.14 MB | 0.76 MB | −33% |

```
ffmpeg -y -i in.gif -movflags faststart -pix_fmt yuv420p \
  -vf "scale=trunc(iw/2)*2:trunc(ih/2)*2" -an out.mp4
```

All **30 live references across 24 files** (v1/v2/v3/v4) were swapped from ``/`` to a `<video autoPlay loop muted playsInline>` element. Each carries **explicit `width`/`height`** (so the browser reserves layout space — a CLS win alongside the LCP one) plus an `aria-label` describing the clip, preserving the alt text the GIFs had. Autoplaying muted inline video keeps the existing "animated screenshot" behaviour on both desktop and mobile Safari.

> `static/gif/record-tc.gif` (278 KB, 11 references) is intentionally **left as a GIF** — it's an order of magnitude smaller than the four above, so the conversion churn isn't worth it in this PR. Noted as an optional follow-up.

**D4 · Analytics loading**
**GA and the Meta Pixel fire eagerly** — GA via the standard `gtag` preset (auto SPA tracking), the Pixel via an inline `headTags` snippet (`init` + `PageView` on load, `<noscript>` fallback intact). This is a **product requirement** (accurate analytics/conversion from first paint), so they are *not* deferred — the measured LCP cost is shown in the table above. **Clarity + Apollo** load on the **first user interaction** (engaged sessions only), and **Hotjar was removed** (redundant session-recorder). `src/metaPixelRouteTracker.js` re-fires the Pixel `PageView` on SPA route changes and drives the Clarity/Apollo interaction gate.

### Performance — CLS

**D5 · Intrinsic `width`/`height` on raw `` tags**
Docusaurus's mdx-loader already resolves and sizes Markdown images (`` renders with `width`/`height` + a content-hashed asset), but it leaves **hand-written HTML `` tags** in `.md`/`.mdx` untouched — and those are the remaining source of layout shift. A new **dependency-free** remark plugin (`src/remark/remarkImageSize.js`) stamps each raw `` with the image's intrinsic `width`/`height`, giving the browser an aspect ratio to reserve space. The global `img { max-width:100%; height:auto }` rule keeps images fully responsive — the attributes only supply the ratio, not a fixed size.

*Why no dependency:* the repo carries **both** `yarn.lock` and `package-lock.json` (Vercel uses yarn, CI uses npm), so `image-size`/`rehype-img-size` would mean keeping two lockfiles in sync. The plugin reads PNG/GIF/JPEG headers itself. The parser is chosen by **magic bytes, not extension**, so a mislabeled file (this repo has one PNG saved as `.jpg`) is still sized correctly — verified against `sips` across **all 188** raster assets (**188/188 exact match**). It never throws, skips remote/relative/`data:`/webp/svg, and never overwrites author-provided dimensions.

Build-verified: **30** raw `` gain dimensions, **69** with author-set widths are left alone, Markdown images are unaffected, **zero** duplicate attributes. No `.md`/`.mdx` source files change — sizing happens at build time.

### Source maps

**D11 · Client source maps**
A `configureWebpack` plugin sets `devtool: 'source-map'` for the client bundle only — no new dependency, emits `.map` files for debuggable first-party JS, zero runtime impact (428 maps emitted).

### Accessibility

**D9 · Decorative SVGs** (+ **X4**)
Added `aria-hidden="true"` to the 5 footer social icons (their `<a>` already carries `aria-label`) and 12 decorative component icons. `DocItem` theme SVGs already had it.

**D10 · Remove dead font reference**
Removed the `"Aeonik"` `font-family` — referenced for headings but never defined via `@font-face`, so it always fell through to the system stack.

### Process / CI

**X1 + X3 · Asset & font budget guard**
New `.github/workflows/asset-budget.yml`, **scoped to files changed in the PR**:
- **X1** — fails on any newly added/modified image > 500 KB.
- **X3** — fails on any newly added render-blocking Google Fonts stylesheet reference (`fonts.googleapis.com/css…`), enforcing D1.

PR-scoped by design, so pre-existing large assets never fail an unrelated PR — only new regressions are caught.

**H1 · SEO sanity** — verified `robots.txt`, `sitemap.xml`, canonical links, and JSON-LD are all present/valid; no change needed.

---

## 🔎 Review feedback addressed

- **CSP has no reporting endpoint** — description corrected to state violations are console-only for now; a `report-to` collector should be wired before enforcing (no fake endpoint added).

## 🔧 Keeping the linters happy

The D3 `<video>` markup tripped two checks; both are fixed in this PR:

- **prettier** reformatted the markdown HTML block because of the indented `<source>` child. The two versions disagree on *how*: 2.8.8 wants a blank line after the opening tag (which would split the JSX block), while 3.9.6 wants the child dedented to column 0. The workflow pins 2.8.8 but the action resolves to 3.9.6, so rather than target either, each `<video>/<source>/</video>` trio is collapsed onto **one line** — verified clean under **both** versions.
- **Vale** flagged `autoPlay` and `playsInline` as misspellings. They're JSX attribute names, not prose, so they're added to the Base vocabulary alongside the other camelCase identifiers already accepted there (`borderRadius`, `containerName`, `matchLabels`, …).

---

## ❌ Not in this PR (and why)

| Ticket | Priority | Why |
|---|---|---|
| **H2** — Product/review schema | P3 `SEO` | **Intentionally not done.** The "product" half is already covered — the site ships a complete `SoftwareApplication` JSON-LD block (the correct schema.org type for a dev tool; a separate `Product` type would be redundant/conflicting). The "review" half is deliberately omitted: `aggregateRating`/`Review` markup for one's own product on one's own domain violates [Google's structured-data policy](https://developers.google.com/search/docs/appearance/structured-data/review-snippet#guidelines) (self-serving reviews) and can trigger a manual action suppressing *all* rich results. Real ratings are surfaced the correct way — via `sameAs` links to G2/Gartner/Capterra/AWS Marketplace in the Organization schema. |
| **D8** — HSTS response header | P2 `SEC` | **Preview-only — committed but not effective in prod.** `vercel.json` sets `Strict-Transport-Security`, but production `keploy.io/docs/*` is served from **S3 + CloudFront**, so `vercel.json` only reaches the **Vercel deploy preview**. The apex domain already sends the identical HSTS value, so prod is covered; enforcing it on the docs path belongs in the CloudFront config, out of this repo. Kept in `vercel.json` (harmless, covers the preview). |
| **X6** — COOP (`Cross-Origin-Opener-Policy`) | P2 `SEC` | Same constraint — the `same-origin` COOP header is in `vercel.json` and therefore **preview-only**; real prod enforcement is a CloudFront-config change. |
| **X7** — CSP + Trusted Types (report-only) | P2 `SEC` | Same constraint — `Content-Security-Policy-Report-Only` (incl. `require-trusted-types-for 'script'`) is **preview-only** and report-only by design (console violations, no enforcement, no `report-to` collector wired). Enforcing CSP/Trusted Types in prod needs the CloudFront config plus a reporting endpoint. |

---

## ✅ Verification

- `npm run build` → **`[SUCCESS]`**; `onBrokenLinks: "throw"` passes — but that validates *links*, not `/<video>/<source>` `src`. That the D2 deletions and D3 swaps break zero references was verified by grep, and is now **enforced** by the asset-budget guard (new step: every added `src="/docs/…"` must resolve under `static/`).
- **CI:** Vale, prettier, asset-budget, run-lint, deploy-preview green. **DCO pending** — sign-off still needed on a few commits.
- **Fonts:** both woff2 subsets emitted + referenced by built CSS (no broken URL); no Google Fonts stylesheet; no `gstatic`/`googleapis` preconnect; 4 preconnects total.
- **D3:** all 4 GIFs deleted with **zero remaining live references** (the only mentions left are inside non-rendering `[//]: #` markdown comments that predate this PR); 30 `<video>` elements across 24 files, each with `width`, `height`, and `aria-label`; MP4s total 1.53 MiB.
- **D5:** remark parser output matches `sips` on **188/188** raster assets; built HTML shows 30 raw `` with injected `width`/`height`, 69 author-sized tags untouched, Markdown images unaffected, 0 duplicate attributes; disabling the plugin drops the attributes (isolation confirmed).
- **D4:** GA eager (gtag preset); Meta Pixel eager (inline `headTags` snippet, `init` + `PageView`, `<noscript>` intact); Clarity + Apollo load on first interaction; Hotjar removed; SPA `PageView` re-fired from the client module.
- **D9 / X4:** 5 footer + 12 component SVGs carry `aria-hidden` (all 16 `<svg>` in `src/` covered, per-tag verified) **plus the 6 inline decorative SVGs in `gsoc/contribution-guide.md`** — the one served-markdown gap the earlier sweep missed.
- **D10:** no real `Aeonik` font-family remains.
- **D11:** 428 `.js.map` files emitted.
- **X1/X3:** workflow valid YAML; simulated on this PR's diff → passes; on a real violation → fails as intended. Image budget now uses `--diff-filter=AMR` so a renamed+re-encoded file can't slip it; added a step that fails on an added `src="/docs/…"` not resolving under `static/`.
- **H1:** `robots.txt`, `sitemap.xml`, canonical, JSON-LD all present/valid.
- `prettier --check` clean on all changed files under both 2.8.8 and 3.9.6; Vale clean on changed lines; no `yarn.lock` churn.

> **Local preview:** `npm run serve` (production preview) can't serve the fonts due to the repo's `trailingSlash: true` setting (it 302→404s every woff2 — the pre-existing Roboto font too), so DM Sans falls back to a system font locally. Use `npm start` (dev server) to preview fonts correctly. On Vercel the self-hosted fonts load fine, exactly like Roboto does today.

---

## 🔄 Review iteration 5 — fixes pushed

- **D9 / X4 gap closed** — the 6 inline decorative GitHub-icon `<svg>` in
  `gsoc/contribution-guide.md` (a served v4 page) now carry
  `aria-hidden="true"` + `focusable="false"`. That was the only served-markdown
  SVG the earlier component/footer sweep missed, so X4 is now genuinely
  complete.
- **X1 rename gap** — the image-budget check used `--diff-filter=AM`, so a
  `git mv` that re-encoded a file larger reported as `R` and was skipped
  entirely. Now `--diff-filter=AMR`.
- **Asset-`src` resolution now guarded** — added an asset-budget step that fails
  when an added `src="/docs/…"` in changed markup doesn't resolve to a file
  under `static/`. `onBrokenLinks` only validates links, so this makes the
  D2/D3 "breaks zero references" claim self-enforcing (and the verification
  line is corrected to stop attributing that proof to the build).
- **D8 / X6 / X7 documented, not dropped** — moved into **❌ Not in this PR** with
  the real reason: those headers (`HSTS`, `COOP`, `CSP-Report-Only`) live in
  `vercel.json`, which only reaches the **Vercel deploy preview** — production
  `/docs/*` is served from S3 + CloudFront, and the apex already sends the same
  HSTS. Prod enforcement belongs in the CloudFront config.

**Deferred (non-blocking, called out honestly):** `concepts/what-are-keploy-features.md`
is the only v4 page with two eager autoplay `<video>`s (~742 KB fetched on load
regardless of viewport). It's the tail of D3, not systemic; a proper fix is an
IntersectionObserver that assigns `src` to the below-fold video. Left for a
follow-up rather than adding a one-page lazy-video component to this PR. Same for
`prefers-reduced-motion` on the autoplay set.


vercel.json
```diff
@@ -3,23 +3,31 @@
...
+        },
+        {
+          "key": "Strict-Transport-Security",
+          "value": "max-age=63072000; includeSubDomains; preload"
+        },
+        {"key": "Cross-Origin-Opener-Policy", "value": "same-origin"},
+        {
+          "key": "Content-Security-Policy-Report-Only",
+          "value": "default-src 'self'; base-uri 'self'; object-src 'none'; frame-ancestors 'self'; script-src 'self' 'unsafe-inline' https://www.googletagmanager.com https://www.google-analytics.com https://connect.facebook.net https://www.clarity.ms https://*.clarity.ms https://telemetry.keploy.io https://assets.apollo.io https://*.algolia.net https://*.algolianet.com; style-src 'self' 'unsafe-inline'; font-src 'self' data:; img-src 'self' data: https:; connect-src 'self' https://www.google-analytics.com https://*.google-analytics.com https://www.googletagmanager.com https://connect.facebook.net https://www.facebook.com https://*.clarity.ms https://telemetry.keploy.io https://keploy.io https://assets.apollo.io https://*.algolia.net https://*.algolianet.com; frame-src 'self' https://www.youtube.com https://www.youtube-nocookie.com; upgrade-insecure-requests"
+        }
```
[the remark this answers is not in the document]
dhananjay6561: Valid catch — thanks. You're right that without a `report-to`/`report-uri` directive (and a matching `Reporting-Endpoints`/`Report-To` header) the browser only logs violations to the console and doesn't transmit them to a collector.

I've corrected the PR description to say violations are **console-only** for now rather than "collected." I intentionally did not add a `report-to` pointing at a non-existent endpoint. When a CSP report collector is available (Sentry / report-uri.com / a Keploy endpoint), we'll wire `Reporting-Endpoints` + `report-to` before flipping the header from Report-Only to enforced.

src/metaPixelRouteTracker.js
```diff
@@ -1,11 +1,70 @@
...
+// without requestIdleCallback (e.g. Safari).
+function scheduleBootstrap() {
+  if (typeof window === "undefined") return;
+  const ric =
+    window.requestIdleCallback ||
+    function (cb) {
+      return window.setTimeout(cb, 2000);
+    };
+  ric(bootstrapPixel);
+}
```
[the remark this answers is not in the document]
dhananjay6561: Fixed in 8f15f65a. Added a module-level `bootstrapScheduled` flag so `scheduleBootstrap()` only ever queues the idle callback once, even across rapid SPA navigations before the first idle fires:

```js
let bootstrapScheduled = false;
function scheduleBootstrap() {
  if (typeof window === "undefined" || bootstrapScheduled) return;
  bootstrapScheduled = true;
  const ric = window.requestIdleCallback || function (cb) { return window.setTimeout(cb, 2000); };
  ric(bootstrapPixel);
}
```

(`bootstrapPixel()` already no-ops on `window.fbq`, so this was never a correctness bug — the flag just avoids redundant timers.)

src/metaPixelRouteTracker.js
```diff
@@ -1,11 +1,76 @@
...
+    scheduleBootstrap();
+    return;
+  }
+  if (location.pathname !== previousLocation.pathname) {
     if (typeof window.fbq === "function") {
       window.fbq("track", "PageView");
+    } else {
+      // Pixel hasn't finished its idle bootstrap yet — schedule it (the stub
+      // will queue this navigation's PageView once fbevents.js loads).
+      scheduleBootstrap();
```
dhananjay6561: 🟡 [important] This comment doesn't match the behavior. If the initial load already called `scheduleBootstrap()` (setting `bootstrapScheduled = true`) but the idle callback hasn't fired yet, and the user navigates before then, this branch calls `scheduleBootstrap()` — which early-returns because `bootstrapScheduled` is already `true`. At that point `window.fbq` doesn't exist yet (the stub is only created inside `bootstrapPixel`), so there is no queue to catch this navigation's PageView. The result is a silently dropped PageView during the idle window (up to the 2s Safari fallback, longer if the main thread is busy), and only the landing page's PageView ever fires.

What happens if the user clicks through 2-3 pages before the browser goes idle? Consider either bootstrapping immediately on the first real SPA navigation (the user is clearly active, so deferral no longer buys much), or recording a `pendingPageView` path that `bootstrapPixel` flushes once the stub exists. At minimum, the comment should be corrected — as written it claims a queue that isn't there yet.

src/remark/remarkImageSize.js
```diff
@@ -0,0 +1,239 @@
...
+  clean = clean.replace(/^\/+/, "");
+
+  const abs = path.join(STATIC_DIR, clean);
+  // Guard against path traversal escaping static/.
+  if (!abs.startsWith(STATIC_DIR + path.sep)) return null;
+  return abs;
+}
+
+// Full pipeline: src string -> {width, height} or null. Never throws.
+const sizeCache = new Map();
```
dhananjay6561, line 166: 🟢 [nit] `sizeCache` is a module-level `Map` that lives for the whole process. In the dev server (`npm start`), if an author replaces an image with one of different dimensions, the stale entry keyed by absolute path sticks until the process restarts, so the injected `width`/`height` go stale. Build-time (`npm run build`) is unaffected since it's a fresh process each time. Not worth blocking on, but a one-line note that the cache is per-process (dev requires a restart to pick up resized assets) would save a future head-scratch.

.github/workflows/asset-budget.yml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: New/changed images must be under 500 KB (X1)
+        run: |
+          base="origin/${{ github.base_ref }}"
+          git fetch origin "${{ github.base_ref }}" --depth=1 || true
```
dhananjay6561: 🟢 [nit] `actions/checkout` already runs with `fetch-depth: 0` (full history), so `origin/<base>` is present with its full ancestry. This extra `git fetch … --depth=1` is redundant, and a shallow fetch layered on top can truncate the ancestry that the three-dot `"$base"...HEAD` merge-base needs — `|| true` would then hide the failure and the diff could come back empty (silently passing the guard). Since the checkout already has everything, I'd just drop this line.

.github/workflows/asset-budget.yml
```diff
@@ -0,0 +1,55 @@
...
+          fail=0
+          while IFS= read -r f; do
+            [ -z "$f" ] && continue
+            [ -f "$f" ] || continue # skip deletions
+            size=$(wc -c < "$f")
+            if [ "$size" -gt "$max" ]; then
+              echo "::error file=$f::$f is $((size / 1024)) KB (> 500 KB). Optimize (WebP/AVIF) or convert to video before committing."
+              fail=1
+            fi
+          done < <(git diff --name-only --diff-filter=AM "$base"...HEAD | grep -iE '\.(png|jpe?g|gif|webp|svg)$' || true)
```
dhananjay6561: 💡 [suggestion] The extension filter matches `png|jpe?g|gif|webp|svg` but not `avif` — yet the failure message on line 36 recommends AVIF as the fix. So a newly committed oversized `.avif` would sail past the budget it's meant to enforce. Adding `avif` (and possibly `mp4|webm` if you ever want a video ceiling) to the regex would close that gap.

dhananjay6561 (commented):
## Code review — perf/web-vitals-docs

Reviewed following the four-phase process (context → high-level → line-by-line → summary). This is a well-scoped, well-documented PR: the ticket table maps every change to a rationale, the remark plugin is genuinely careful (magic-byte parsing, never-throws, path-traversal guard, author-dimension preservation), and the risky bits (CSP, source maps) are correctly landed in non-enforcing / no-runtime-impact modes. Build + CI are green and the description is honest about what's deferred and why.

No blocking issues. I left a handful of inline notes — one 🟡 (a real but narrow analytics gap in the pixel loader whose code comment overstates what happens), plus a few 🟢/💡 hardening nits on the plugin cache and the CI guard. None need to block merge; the pixel one is worth a quick look.

🎉 The `remarkImageSize` plugin is the standout — choosing the parser by magic bytes rather than extension (and verifying 188/188 against `sips`) is exactly the right call for a repo with a mislabeled asset.

src/metaPixelRouteTracker.js
dhananjay6561: Good catch — fixed in 3f09246e. On a real SPA navigation before the idle bootstrap has fired, we now call `bootstrapPixel()` synchronously instead of re-scheduling into the not-yet-fired idle window. That creates the `fbq` stub and fires `init` + this page's `PageView` immediately (the stub queues them until `fbevents.js` loads), so the navigation is no longer dropped. Any already-scheduled idle callback then no-ops because `window.fbq` exists. Comment updated to describe the actual behavior.

src/remark/remarkImageSize.js
dhananjay6561: Added a note above `sizeCache` in 3f09246e: the cache is module-level and persists for the whole process, so `npm run build` is always accurate (fresh process) while the `npm start` dev server needs a restart to pick up resized assets.

.github/workflows/asset-budget.yml
dhananjay6561: Dropped in 3f09246e. Since `actions/checkout` already runs with `fetch-depth: 0`, `origin/<base>` is present with full ancestry, so the extra shallow fetch was redundant (and risked truncating the ancestry the three-dot merge-base needs). Added a short comment noting why it's not needed.

.github/workflows/asset-budget.yml
dhananjay6561: Added `avif` to the extension filter in 3f09246e, so an oversized `.avif` is now caught by the same 500 KB budget the failure message recommends AVIF for. Left `mp4`/`webm` out on purpose — the image-budget message tells authors to *convert to video* as the escape hatch, so capping video here would contradict that advice.

dhananjay6561 (commented):
## Re-review — all four notes resolved ✅

Re-reviewed the three follow-up commits (`5e32dce`, `256c6a6`, `3f09246`) against my earlier comments. Each is fixed correctly:

- 🟡 **Dropped PageView** (`5e32dce`) — first real SPA navigation now calls `bootstrapPixel()` synchronously instead of re-scheduling into the not-yet-fired idle window. That creates the `fbq` stub and fires `init` + this page's `PageView` (queued until `fbevents.js` loads), so the navigation is no longer lost, and any already-scheduled idle callback cleanly no-ops via the `window.fbq` guard. The corrected comment now matches the behavior. Resolved.
- 🟢 **`sizeCache`** (`256c6a6`) — per-process caveat documented (build = fresh process/accurate; dev server = restart to pick up resized assets). Resolved.
- 🟢 **Redundant shallow fetch** (`3f09246`) — dropped, with a comment noting `fetch-depth: 0` already provides the ancestry the three-dot merge-base needs. Resolved.
- 💡 **AVIF budget gap** (`3f09246`) — `avif` added to the image regex, so the guard now matches the format its own error message recommends. Resolved.

No new issues introduced. Nothing outstanding from my side — LGTM. 🎉

amaan-bhati:
# Claude Review Skill, Iteration 3

**Scope:** the 4 commits landed after Iteration 2 (`34a6383`, `814db2d`, `9d68f6b`, `2295ef4`) plus a full re-pass over the whole diff (57 files, +537 / -112). Four-phase process: context, high level, line by line, summary.

**VERDICT: 🔄 REQUEST CHANGES** (1 blocking). No code here is wrong. The blocker is that 3 tickets marked ✅ Done ship nothing to real users.

**Tally:** 🔴 1 · 🟡 4 · 🟢 4 · 💡 2 · 📚 1 · 🎉 5

---

## 🔴 BLOCKING (1)

### 🔴 D8 / X6 / X7 never reach production. `keploy.io/docs/*` is served from S3, not Vercel.

Measured on live prod:

```
$ curl -sI https://keploy.io/docs/keploy-explained/introduction/
HTTP/2 200
server: AmazonS3
via: 1.1 ... (CloudFront)
cache-control: max-age=public,
```

No `x-frame-options`, no `x-content-type-options`, no `referrer-policy`, no `permissions-policy`. Those four are already in `vercel.json` on `main` and are absent in prod, which proves `vercel.json` is not the serving path. `.github/workflows/main.yml:45-56` builds and pushes to S3 via `reggionick/s3-deploy` plus a CloudFront invalidation. `vercel.json` governs the deploy preview only.

So HSTS, COOP, CSP-Report-Only and Trusted Types land nowhere for real users.

Two related facts:

* The apex already sends the identical HSTS value, and HSTS is per host, so `/docs` is covered:
  `strict-transport-security: max-age=63072000; includeSubDomains; preload` on `https://keploy.io/`. The HSTS line in this PR is redundant even if it did ship.
* The apex enforces a full CSP (with `worker-src`, `form-action`, `media-src`). Docs pages get **no CSP at all**. That is the real gap, and `vercel.json` does not close it.

**Fix:** deliver these via a CloudFront response headers policy (or S3 object metadata in the deploy step), then re-label D8 / X6 / X7 in the ticket table. Keeping `vercel.json` for previews is fine, just do not mark them done.

---

## 🟡 IMPORTANT (4)

**🟡 1. `Cache-Control` on prod is malformed, so D1 and D3 lose their repeat-visit win.**
`main.yml:54` sets `cache: "public, max-age:86400"`, a colon instead of `=`. Prod serves `cache-control: max-age=public,`, an invalid delta-seconds, so nothing under `static/` gets a valid freshness lifetime. The new woff2 pair (54 KB), 4 MP4s (1.53 MB) and 4 posters (228 KB) all inherit it. Google Fonts previously served DM Sans with a long immutable cache, so self-hosting is a repeat-visit regression until this is fixed. One character, plus ideally a `max-age=31536000, immutable` rule for `/docs/fonts/`.

**🟡 2. Stale comment contradicts the shipped strategy.**
`docusaurus.config.js:641-646` says `GA + Meta Pixel -> on idle` and "loaded from src/metaPixelRouteTracker.js instead". Shipped state: GA via the gtag preset, Pixel via the `headTags` inline snippet, both eager. Only Clarity and Apollo moved into the client module. This is the third place the strategy is written down and the only one that is wrong.

**🟡 3. `mousemove` defeats the Clarity/Apollo interaction gate on desktop.**
`src/metaPixelRouteTracker.js:24`. Any desktop visitor moves the pointer within milliseconds of paint, so both scripts load almost immediately. "Engaged sessions only" holds for `touchstart` and `scroll`, not for `mousemove`. Drop it, or put it behind a movement threshold.

**🟡 4. 12 of 30 `<video>` tags hardcode `https://keploy.io`.**
All in the v1 to v4 glossary files, for both `poster` and `<source>`. Inherited from the old ``, but it matters more now: the deploy preview and localhost pull **production** media, so the new MP4s are not verifiable in the preview for 12 of 30 spots. The docs CSP also declares no `media-src`, so it falls back to `default-src 'self'`, which passes on the prod origin by accident and would block anywhere else once enforced. Make them relative `/docs/...` like the other 18.

---

## 🟢 NITS (4)

* **No pause control on 30 autoplay looping videos.** WCAG 2.2.2. Not a regression (GIFs were worse), but `<video>` makes it fixable: gate autoplay on `prefers-reduced-motion` in JS, since CSS cannot stop autoplay.
* **`record-api.mp4` is 744 KB, 2.7x the next largest, and oversized.** 1074x782 intrinsic, rendered at 80% of a ~750px column. Its poster is 154 KB, also the largest of the four. Downscaling to ~800px wide cuts both.
* **X3 font guard misses `.css`.** `asset-budget.yml:49` pathspec has no `*.css`, so `@import url(https://fonts.googleapis.com/css2...)` in `custom.css` slips past. That is exactly the mechanism the old code comment referenced.
* **Docs CSP diverges from the apex policy.** Missing `form-action` and `media-src` (apex has both), and `script-src 'unsafe-inline'` with no nonce buys little XSS protection while `require-trusted-types-for 'script'` is declared. Two divergent policies on one origin is a maintenance trap.

---

## 💡 SUGGESTIONS (2)

* **Follow the Roboto precedent instead of an inline `<style>`.** `custom.css:206-228` already does `@font-face { src: url("../fonts/Roboto-Light.woff2") }` from `src/fonts/`, webpack-processed and content-hashed. The new comment claims css-loader "refuses to resolve" the URL from `src/css`, but Roboto proves relative paths resolve fine there. That route keeps `@font-face` in CSS, drops the inline `<style>` (which needs `style-src 'unsafe-inline'`), and gets content hashing plus immutable caching for free, which also sidesteps 🟡1. Minor: the PR table lists `src/fonts/DMSans-*.woff2`, the files actually landed in `static/fonts/`.
* **D2 missed dead JS while sweeping dead images.** `static/js/code-block-buttons.js` has zero references. `static/scripts/fullstory.js` and `chat.js` are referenced only from the commented block at `docusaurus.config.js:654-668`. Same ticket, one more pass.

---

## 📚 LEARNING (1)

`remarkImageSize.resolveStaticPath` skips paths starting with `.` as relative, but a bare `img/x.png` (relative, no dot) is treated as static-rooted, so it could stamp another file's dimensions if a same-named file exists under `static/`. Zero such tags in the repo today, so latent only.

---

## 🎉 PRAISE (verified, not taken on trust)

* Magic-byte parser selection plus the 188/188 `sips` cross-check. Right call for a repo with a mislabeled asset.
* All 4 poster/MP4 pairs match exactly, checked with `ffprobe`: 800x348, 800x450, 562x356, 1074x782. Every declared `width`/`height` matches the real stream, so the reserved space is correct. Posters are paired to the right source in all 30 tags.
* `img, video { max-width: 100%; height: auto }` does exist at `custom.css:933-937`, so the injected attributes supply ratio without fixing size. Claim holds.
* Zero live references to the 10 deleted assets. The 3 remaining mentions sit inside non-rendering `[//]: #` comments, exactly as stated.
* Honest "Not in this PR" section, and the call on self-serving review schema is correct.

---

## Verify before merge

1. Confirm where response headers for `keploy.io/docs/*` actually originate, then move D8 / X6 / X7 there.
2. `curl -I` the woff2 on the deploy preview and check both the 200 and the `Cache-Control`.
3. `versioned_docs/version-3.0.0` is excluded by `onlyIncludeVersions` (`docusaurus.config.js:461`), so the 5 v3 files edited here are never compiled. Not build-verified.

---

*Reviewed with the [code-review-skill](https://github.com/awesome-skills/code-review-skill) four-phase process. Severity: 🔴 blocking · 🟡 important · 🟢 nit · 💡 suggestion · 📚 learning · 🎉 praise.*

amaan-bhati:
# Claude Review Skill, Iteration 4

**Scope:** commit `944bbd33` ("address review iteration 3"), re-verified against every finding from Iteration 3. Diff vs the previously reviewed head: 23 files, +19 / -136.

**VERDICT: 💬 COMMENT.** Code is ready to merge. Eight of nine findings are fixed correctly and I confirmed each one rather than taking the commit message for it. One item is left: it is a description fix, not a code fix.

**Tally:** 🔴 0 · 🟡 1 (description only) · 🟢 2 open · ✅ 8 fixed

---

## ✅ FIXED AND VERIFIED (8)

| Iter-3 finding | Fix | Verified how |
|---|---|---|
| 🟡 Malformed `Cache-Control` | `main.yml:54` now `public, max-age=86400` | `=` replaces `:` |
| 🟡 Stale analytics comment | Rewritten to name the real mechanism per tracker | `docusaurus.config.js:641-647` matches the shipped config |
| 🟡 `mousemove` defeats the gate | Dropped, with a comment saying why | `INTERACTION_EVENTS` is now 4 events |
| 🟡 12 hardcoded `https://keploy.io` videos | All 12 relative | 0 absolute URLs left in any `<video>` tag |
| 🟢 `record-api.mp4` oversized | 744 KB → **318 KB** (-57%), poster 154 KB → **56 KB** (-63%) | `ffprobe`: stream is 800x582, poster is 800x582, declared `width`/`height` is 800x582. All three agree |
| 🟢 X3 guard misses `.css` | `*.css` added to the pathspec | `asset-budget.yml:49` |
| 📚 Bare relative `` | Now requires a leading `/`, so `img/x.png` is skipped | `resolveStaticPath`. Stricter than I suggested, and correct |
| 💡 Dead JS missed by D2 | `code-block-buttons.js`, `chat.js`, `fullstory.js` deleted (125 lines) plus the commented config block | file list |

CI is green on all 7 checks.

---

## 🟡 STILL OPEN (1, description only)

**🟡 D8 / X6 / X7 were deleted from the ticket table instead of being re-labeled.**

The rows are gone from the status table, but:

* `vercel.json` still ships HSTS + COOP + CSP-Report-Only in this PR.
* The three tickets appear nowhere in **"❌ Not in this PR (and why)"**.

The PR summary says: *"Genuinely-blocked tickets (and why) are listed at the bottom so nothing is silently dropped."* These are now silently dropped, which is the one outcome that sentence rules out. Deleting the row also loses the finding: the next person to pick up D8 has no record that `vercel.json` cannot deliver it.

**Fix:** move D8 / X6 / X7 into the "Not in this PR" table with the reason (production `/docs/*` is served from S3 + CloudFront, so `vercel.json` only reaches the deploy preview), and note that the apex already sends the identical HSTS value. Keeping the `vercel.json` block is fine; it just needs to be described as preview-only.

---

## 🟢 NOT ADDRESSED, FINE TO DEFER (2)

* **CSP has no `form-action` or `media-src`, and `script-src 'unsafe-inline'` has no nonce.** Unchanged. Only bites when the header is actually delivered, so it belongs with the CloudFront work above rather than here.
* **No `prefers-reduced-motion` handling for the 30 autoplay looping videos.** Unchanged. Not a regression against the GIFs, and needs JS since CSS cannot stop autoplay. Reasonable as a follow-up.

The 💡 about moving the font to `src/fonts/` with the Roboto relative-url pattern is worth much less now that the cache header is fixed. Dropping it is the right call.

---

## 🎉 PRAISE

* Every fix was verified rather than asserted, and the two I expected to be fudged were not: the `record-api` re-encode updated the declared `width`/`height` **and** regenerated the poster at the new size, so the aspect ratio stays exact.
* The `resolveStaticPath` fix is stricter than the finding asked for. Requiring a leading `/` closes the whole class instead of the one case I named.
* The `mousemove` removal carries a comment explaining the reasoning, so it will not get re-added by someone adding "more interaction signals".

---

*Reviewed with the [code-review-skill](https://github.com/awesome-skills/code-review-skill) four-phase process. Severity: 🔴 blocking · 🟡 important · 🟢 nit · 💡 suggestion · 📚 learning · 🎉 praise.*

amaan-bhati:
# Claude Review Skill, Iteration 5

**Scope:** no new commits since Iteration 4 (head is still `944bbd33`), so this is a deeper pass over ground the earlier iterations did not cover: the D9/X4 sweep beyond JSX, the X1 guard's filter semantics, what the build actually validates, and the remaining eager media cost.

**VERDICT: 💬 COMMENT.** Two things before merge, neither is new code. One is the Iteration 3 blocker still unresolved, one is a new gap in D9/X4.

**Tally:** 🔴 1 carried · 🟡 1 new · 🟢 3 new · ✅ 2 risks ruled out

---

## 🟡 NEW: D9 / X4 stop at JSX and miss 6 SVGs in served markdown

All 16 `<svg>` in `src/` carry `aria-hidden`, `aria-label`, or `role`. Clean. But `versioned_docs/version-4.0.0/gsoc/contribution-guide.md` has **6 inline decorative `<svg>` with none of the three**, so they are still exposed to the accessibility tree on a served v4 page.

D9 reads *"`aria-hidden` on 5 footer social + 12 decorative component SVGs ✅ Done"* and X4 (*"accessibility-tree well-formedness for AI crawlers"*) is marked resolved by D9. The sweep was scoped to components and the footer, not to the version. Either add `aria-hidden="true"` to those 6 or mark X4 partial.

---

## 🟢 NEW (3)

**🟢 1. X1 misses renames.** `asset-budget.yml:40` uses `--diff-filter=AM`. Rename detection is on by default in `git diff`, so a renamed file reports as `R` and is skipped entirely. `git mv static/big.png static/img/big.png` while re-encoding it larger slips the 500 KB budget. Use `--diff-filter=AMR`, or drop the filter and let the existing `[ -f "$f" ] || continue` handle deletions.

**🟢 2. The verification line misattributes its own evidence.** The PR says *"`npm run build` → `[SUCCESS]`; `onBrokenLinks: "throw"` passes (the D2 deletions and D3 swaps break zero references)"*. Docusaurus's broken-link checker validates **links**, not `` / `<video>` / `<source>` `src` attributes, so the build cannot prove that claim. The grep did. I checked the conclusion separately and it holds: nothing in `src/` builds an asset path by concatenation or template literal, so there is no dynamic reference the grep could have missed. The claim is correct, the stated proof is not, and nothing guards it going forward. Worth extending `asset-budget.yml` with a check that every `src="/docs/…"` in changed markdown resolves to a file under `static/`. That makes the claim self-enforcing and costs about ten lines.

**🟢 3. One page still ships 742 KB of eager autoplay media.** `concepts/what-are-keploy-features.md` carries 2 videos plus posters. Autoplay forces the fetch, and there is no lazy equivalent for `<video>`, so both download on load regardless of viewport. It is the only 2-video page in v4, so this is the tail end of D3 rather than a systemic issue: an IntersectionObserver that assigns `src` to the below-fold one would clear it.

---

## ✅ RISKS CHECKED AND RULED OUT (2)

* **EXIF orientation in `remarkImageSize`.** A JPEG with orientation 5 to 8 displays with swapped dimensions, so stamping the raw SOF values would produce an inverted aspect ratio and *cause* layout shift instead of fixing it. Their `sips` cross-check would not have caught it either, since `sips` reports stored dimensions too. Checked all 13 JPEGs under `static/`: **0 rotated**. Not an issue, now or by accident.
* **D9 coverage in components.** 16 of 16 `<svg>` in `src/` are covered, and the footer SVGs in `docusaurus.config.js` all carry `aria-hidden` plus `focusable="false"`. The only gap is the markdown one above.

---

## 🔴 CARRIED FROM ITERATION 3, STILL OPEN

**D8 / X6 / X7 were deleted from the ticket table rather than re-labeled.** `vercel.json` still ships HSTS + COOP + CSP-Report-Only, the three tickets appear nowhere in "❌ Not in this PR (and why)", and the summary still promises that blocked tickets are listed there so nothing is silently dropped. Move them into that table with the reason: production `/docs/*` is served from S3 + CloudFront, so `vercel.json` only reaches the deploy preview, and the apex already sends the identical HSTS value.

Also still open and fine to defer: CSP has no `form-action` or `media-src`, and there is no `prefers-reduced-motion` handling for the 30 autoplay videos.

---

## 🎉 PRAISE

The magic-byte parser choice turns out to be load-bearing in the narrowest possible way. The **only** raw `` in v4 with a `.jpg` extension is `/docs/img/keploy-test-openhospital.jpg`, and that file is the mislabeled PNG. An extension-based parser would have handed it to the JPEG reader, failed the SOI check, and silently skipped the one JPEG-named image the plugin actually has to size. Choosing by magic bytes was not defensive programming for a hypothetical; it was required for the single real case.

---

*Reviewed with the [code-review-skill](https://github.com/awesome-skills/code-review-skill) four-phase process. Severity: 🔴 blocking · 🟡 important · 🟢 nit · 💡 suggestion · 📚 learning · 🎉 praise.*

dhananjay6561:
Thanks — the magic-byte praise and the two risks you ruled out (EXIF orientation across the 13 JPEGs, component-SVG coverage) are appreciated. Addressed the actionable items; head is now `97467f3d`.

### 🟡 NEW — D9/X4: 6 decorative SVGs in the GSoC guide
Fixed in `e302944e`. The 6 inline GitHub-icon `<svg>` in `versioned_docs/version-4.0.0/gsoc/contribution-guide.md` now carry `aria-hidden="true"` + `focusable="false"` — that was the one served-markdown gap the component/footer sweep missed, so X4 is now genuinely complete (verified in the built HTML).

### 🟢 1 — X1 misses renames
Fixed in `97467f3d`. `--diff-filter=AM` → `--diff-filter=AMR`, so a `git mv` that re-encodes a file larger can't slip the 500 KB budget.

### 🟢 2 — verification claim vs. its evidence
Fixed in `97467f3d`, and you're right that `onBrokenLinks` proves nothing about `/<video>/<source>` `src`. Added an asset-budget step that fails when an added `src="/docs/…"` in changed markup doesn't resolve under `static/`, so the D2/D3 "breaks zero references" claim is now self-enforcing rather than asserted. Corrected the verification line in the PR body to attribute the proof to the grep + guard, not the build. Both guards simulated against this PR's real base → pass, no false positives.

### 🔴 carried — D8 / X6 / X7 silently dropped from the ticket table
Fixed in the PR description: added D8 (HSTS), X6 (COOP), X7 (CSP + Trusted Types, report-only) to **❌ Not in this PR (and why)** with your reasoning — those headers live in `vercel.json`, which only reaches the Vercel deploy preview; production `/docs/*` is served from S3 + CloudFront, and the apex already sends the identical HSTS. So nothing is silently dropped now. (Kept in `vercel.json` since it's harmless and covers the preview; real prod enforcement belongs in the CloudFront config. Happy to strip them from `vercel.json` instead if you'd prefer the PR carry zero preview-only config.)

### 🟢 3 — one page still ships 742 KB of eager autoplay media
Agreed it's the tail of D3, not systemic (`concepts/what-are-keploy-features.md` is the only 2-video v4 page). Deferred to a follow-up rather than adding a one-page IntersectionObserver lazy-video component to this PR, and called out explicitly in the PR description alongside the `prefers-reduced-motion` gap. Can take it here if you'd rather it not ship as-is.

### ✅ Risks you ruled out
No action — matches what's shipped. Thanks for checking the EXIF/orientation case against the actual JPEGs.

amaan-bhati:
# Claude Review Skill, Iteration 6

**Scope:** the 2 commits after `944bbd33` (`e302944e` aria-hidden, `97467f3d` CI guards) plus the description rework. Diff vs the last reviewed head: 3 files, +37 / -15.

**VERDICT: 🔄 REQUEST CHANGES.** Everything from Iteration 5 is fixed, and the blocker I have carried since Iteration 3 is now properly closed. But **Vale is red, and this PR caused it.**

**Tally:** 🔴 1 new · 🟡 2 new · ✅ 4 fixed

---

## 🔴 BLOCKING: Vale is failing, and the aria-hidden commit is why

`Vale doc linter` reports exactly one result:

```
[Vale.Spelling] Did you really mean 'Autogenerate'?
  versioned_docs/version-4.0.0/gsoc/contribution-guide.md:87
```

Line 87 is `<td>Autogenerate Test Cases</td>`. It is pre-existing and untouched. `e302944e` edited **line 90**, the `<svg>` three lines down. `vale-lint-action.yml` sets `filter_mode: diff_context` with `fail_on_error: true`, so a pre-existing error inside the context window of a changed line becomes a build failure.

Worth being precise about, because it is **not** the failure mode PR #897 has: this PR is 62 files, comfortably under the 100-per-page changed-files limit, so reviewdog's diff scoping is working correctly here. This is a legitimate in-context report, not the whole-tree fallback.

**Fix:** add `[Aa]utogenerate` to `vale_styles/config/vocabularies/Base/accept.txt`. That is the same mechanism PR #897 uses for eleven words, so it is already the house pattern. Rewording the table cell also works.

---

## 🟡 NEW (2), both in the new asset-reference guard

The guard is a good addition and it correctly encodes the `static/` to `/docs/` mapping. Two gaps:

**🟡 1. It checks `src=` but not `poster=`, so half of D3's new references are unguarded.**

This PR adds exactly **30** `src="/docs/…"` refs and exactly **30** `poster="/docs/…"` refs, because every converted `<video>` has both. The regex is `grep -oE 'src="/docs/[^"]+"'`, so all 30 posters are invisible to it. A typo'd poster path 404s silently, the frame stays blank until the video paints, and CI passes. Add `poster` to the alternation:

```
grep -oE '(src|poster)="/docs/[^"]+"' | sed -E 's#^(src|poster)="/docs/##; s#"$##'
```

**🟡 2. The guard cannot support the D2 half of its own name.**

The step is titled `(D2/D3)` and its comment says it exists because *"the 'deletions break zero references' claim needs its own guard here."* It cannot serve that claim: the pipeline is `git diff … | grep -E '^\+'`, so it only ever sees **added** lines. A deletion breaks *pre-existing* references sitting on unchanged lines, which this never reads. It fully guards D3 (new refs must resolve) and nothing of D2.

The D2 direction is the mirror image:

```bash
git diff --diff-filter=D --name-only "$base"...HEAD -- 'static/*'
# for each deleted path, grep the tree for "/docs/<relative path>"
```

Either add that, or retitle the step `(D3)` so the coverage claim matches the code.

---

## ✅ FIXED AND VERIFIED (4)

| Iteration 5 finding | Fix | Verified how |
|---|---|---|
| 🔴 D8 / X6 / X7 deleted from the ticket table | Now rows in **❌ Not in this PR (and why)** | All three present with the S3 + CloudFront reason, that the apex already sends the same HSTS, and that the `vercel.json` block is kept as preview-only. Plus a "documented, not dropped" note |
| 🟡 6 GSoC SVGs missing `aria-hidden` | All 6 have `aria-hidden="true" focusable="false"` | Full re-audit across `src/`, `docusaurus.config.js`, and all of v4 markdown: **27 `<svg>`, 0 missing** an a11y attribute |
| 🟢 X1 misses renames | `--diff-filter=AM` to `AMR` in both steps | Both call sites, with a comment naming the escape it closes |
| 🟢 `onBrokenLinks` credited for asset refs | New guard step added, and the comment now states plainly that `onBrokenLinks` validates links, not `src` | See the two gaps above |

The `smart-set-agent.md` changes in the compare came in via the `main` merge, not from this PR (it is absent from the PR's diff against `main`), so I have not attributed them here.

---

## 🎉 PRAISE

The D8 / X6 / X7 write-up is better than what I asked for. I suggested moving the rows and giving the reason; the entry also records that the headers **are** committed and harmless, that HSTS is already covered at the apex so prod is not actually exposed, and that enforcement belongs in the CloudFront config rather than this repo. That is the version a future maintainer can act on without re-deriving the S3 finding.

---

*Reviewed with the [code-review-skill](https://github.com/awesome-skills/code-review-skill) four-phase process. Severity: 🔴 blocking · 🟡 important · 🟢 nit · 💡 suggestion · 📚 learning · 🎉 praise.*

dhananjay6561:
@amaan-bhati both 🟡 on the asset-reference guard are addressed in `cb2f36bb`.

**1. `poster=` was unguarded** — right, the regex only matched `src=`, so all 30 `<video poster="…">` refs were invisible to it. Added `poster` to the alternation:

```sh
grep -oE '(src|poster)="/docs/[^"]+"' | sed -E 's#^(src|poster)="/docs/##; s#"$##'
```

Simulated on this PR's diff, the guard now checks **30 src + 30 poster** (8 unique paths after `sort -u`), all resolve — and a typo'd poster now fails instead of silently 404-ing behind the video.

**2. The step can't serve the D2 half of its name** — agreed. The pipeline is `git diff … | grep '^\+'`, so it only ever sees added lines: it guards the added-reference (D3) direction, never the deletion (D2) one. I took the **retitle** option rather than adding the deletion scan, because that scan would false-positive on *this very PR*: `record-api.gif` is still named in `[//]: #` comments (with the `/docs/` prefix) in the v2/v3/v4 feature pages — non-rendering, pre-existing. A whole-tree grep would flag those and fail CI, and robustly stripping every comment form (`[//]: #`, `<!-- -->`, `{/* */}`) across md/mdx/js/html is more fragility than the guard is worth. So the step is now titled **(D3)** and its comment states the scope honestly and why the D2 direction is intentionally out.

Thanks for the catch on both.

amaan-bhati (commented):
## Reviewed locally at `218fe219`. Two things before I can approve

**1. This PR removes the Vale doc linter repo-wide, and the stated reason doesn't hold.**

`122949fa ci(docs): remove Vale doc linter` deletes `.vale.ini`, `.github/workflows/vale-lint-action.yml`, and all 37 files under `vale_styles/`. Its reasoning:

> The Vale check fails on pre-existing spelling/style issues in files unrelated to this PR (versioned_docs/version-2.0.0/*), blocking merge.

Vale run history says otherwise — it passes on every other recent PR against the same tree:

```
failure  pull_request  feat/ai-citation-health           2026-08-25
success  pull_request  feat/native-macos-windows-support  2026-08-23
success  pull_request  docs/windows-no-admin              2026-08-22
success  pull_request  docs/mock-your-tests               2026-08-22
success  pull_request  feat/native-macos-windows-support  2026-08-22
```

The action filters to changed lines, so pre-existing `version-2.0.0` errors don't fail anyone. I confirmed this on #897: of **549** Vale errors in the tree, only **2** sit on lines that PR added. Pre-existing errors aren't what blocks a PR.

Two more things on this:

- It isn't in the title or scope (`self-host fonts, drop dead assets, defer pixel, add security headers`), and the body still describes Vale as live and green — *"**Vale** flagged `autoPlay` and `playsInline`… added to the Base vocabulary"* and *"**CI:** Vale, prettier, asset-budget, run-lint, deploy-preview green"*. A reviewer reading the body would think the linter still runs.
- If this merges before #897, that PR's Vale failure disappears without being fixed.

Removing a linter may be a fine call, but it wants its own PR and its own argument.

**2. DCO: 5 commits need sign-off.**

```
97467f3d  ci(docs): catch renamed assets and unresolved /docs src refs (X1)
e302944e  fix(docs): aria-hidden the GSoC guide's decorative SVGs (D9/X4)
3f09246e  ci(asset-budget): drop redundant shallow fetch; add avif to image budget filter
256c6a6c  docs(remark): note sizeCache is per-process (dev restart to pick up resized assets)
5e32dcef  fix(pixel): bootstrap immediately on first SPA nav to avoid dropped PageView
```

18 non-merge commits total, 13 signed.

**What I verified as clean:** all 53 deletions, no dangling references. `static/img/record-api.gif` looked like a live reference in three `what-are-keploy-features.md` files, but each is inside a Markdown comment (`[//]: # ''`), so nothing renders it. Only `.vale.ini` in `README.md` is left over, and that follows finding 1. Everything else green: asset-budget, deploy-preview, prettier, run-lint.

Happy to approve once the Vale removal is split out or justified, and the five commits are signed.

dhananjay6561:
Thanks for the careful review — both addressed.

**1. Vale removal.** This is intentional, not an incidental fix for a failing check — Neha asked us to retire the Vale linter, so it is being removed here deliberately. You are right that the commit message's stated rationale (pre-existing `version-2.0.0` errors blocking merge) does not hold, since the action filters to changed lines — that reasoning was wrong; the real reason is the maintainer decision to drop Vale.

**2. DCO.** Fixed. The five commits are signed off and pushed; all commits now carry `Signed-off-by`. Note I rebased the branch linearly onto `main` (the two "Merge main" commits are gone and hashes changed), but the tree is byte-identical to what you reviewed at `218fe219`, so your content review still holds.

amaan-bhati (commented):
## Reviewed locally at `048a55f9`. One thing before I approve

DCO is fixed - all 18 non-merge commits signed. prettier clean on all 37 changed files. Verified the deletions too: of 53 removed files, nothing live still references them.

**The Vale removal is the blocker, and the reason given for it doesn't hold.**

`13827a1a ci(docs): remove Vale doc linter` deletes `.vale.ini`, the workflow, and all 37 `vale_styles/` files. Its rationale:

> The Vale check fails on pre-existing spelling/style issues in files unrelated to this PR (versioned_docs/version-2.0.0/*), blocking merge.

Run history for that workflow says otherwise — same tree, same pre-existing errors:

```
failure  feat/ai-citation-health            2026-08-26
failure  feat/ai-citation-health            2026-08-25
success  feat/native-macos-windows-support  2026-08-23
success  feat/native-macos-windows-support  2026-08-23
success  feat/native-macos-windows-support  2026-08-23
success  docs/windows-no-admin              2026-08-22
success  docs/mock-your-tests               2026-08-22
success  feat/native-macos-windows-support  2026-08-22
success  docs/mock-your-tests               2026-08-22
success  docs/mock-your-tests               2026-08-22
```

The action filters to changed lines, so pre-existing `version-2.0.0` errors don't block anyone. It fails on exactly one branch — #897 — and there on **2 errors out of 549**, both the word `sanitization` on one added line. A one-word vocabulary entry fixes that.

Two more things worth knowing:

- **The PR contradicts itself.** Commit 5 (`0a7cb8db`) adds `autoPlay` and `playsInline` to `vale_styles/config/vocabularies/Base/accept.txt`; commit 17 (`13827a1a`) deletes all 38 files under `vale_styles/`. That vocabulary entry doesn't exist at head — I checked. The removal reads as a late reaction, not a decision.
- **It isn't in the title or scope** (`self-host fonts, drop dead assets, defer pixel, add security headers`), and nothing in the body says the linter is going away.

Dropping Vale may well be the right call, but it wants its own PR and its own argument rather than riding along inside a perf change.

**Everything else verified clean:**

```
DCO                      18/18 non-merge commits signed
prettier 2.8.8           37 changed files, all pass --check
53 deletions             no live dangling references
53 rooted media refs     52 resolve on disk
```

The two apparent reference hits are both benign: `record-api.gif` appears only inside Markdown comments (`[//]: # ''`, three files, nothing renders it), and `.vale.ini` in `README.md` follows from the removal above.

One pre-existing issue I noticed, out of scope: `/docs/img/Keploy-record-openhospital.png` is referenced by `versioned_docs/version-2.0.0/quickstart/java-spring-boot-openhospital.md:79` but doesn't exist on `main` either. Your new D3 guard reads only added diff lines, so it correctly doesn't flag it - a full-tree sweep would be a reasonable follow-up.

Happy to approve once the Vale removal is split out or argued on its own terms.

amaan-bhati (approved):
## Approving. Verified locally at `048a55f9` (but needs to address the usage of jpg images)

**Verified your rebase claim rather than taking it:**

```
git diff 218fe219 pr-896   ->  0 lines
tree 218fe219 = 4cd0e3fa2ca5
tree 048a55f9 = 4cd0e3fa2ca5   => identical
merge commits: 0   non-merge: 18
```

Byte-identical tree, so the content review from the earlier head carries over intact and the history is linear.

**DCO fixed:**

```
18/18 non-merge commits carry Signed-off-by
```

**Re-verified locally at this head:**

```
prettier 2.8.8        37 changed files, all pass --check
53 deletions          no live dangling references
53 rooted media refs  52 resolve under static/ (baseUrl /docs/)
CI                    DCO, asset-budget, deploy-preview, greeting, prettier, run-lint all pass
```

The two apparent reference hits are both benign, as your description already notes: `record-api.gif` survives only inside non-rendering `[//]: #` Markdown comments in three files, and `.vale.ini` in `README.md` follows from the removal.

## 🟢 Two documentation nits, non-blocking

The description hasn't caught up with the branch:

- Line 163: *"**CI:** Vale, prettier, asset-budget, run-lint, deploy-preview green. **DCO pending** - sign-off still needed on a few commits."* Vale no longer runs, and DCO is now green.
- Line 145 describes adding `autoPlay`/`playsInline` to the Base vocabulary as a live fix. That was commit 5 (`0a7cb8db`); commit 17 (`13827a1a`) deletes all 38 `vale_styles/` files, so the entry doesn't exist at head — I checked.

Worth a pass so someone reading this in six months doesn't go looking for a linter that isn't there. `13827a1a`'s message also still carries the rationale you agreed doesn't hold, which is permanent in history - a follow-up commit noting the real reason would cover it.

One pre-existing issue, out of scope: `/docs/img/Keploy-record-openhospital.png` is referenced by `versioned_docs/version-2.0.0/quickstart/java-spring-boot-openhospital.md:79` but exists on neither branch. Your D3 guard reads only added diff lines so it correctly ignores it; a full-tree sweep would be a reasonable follow-up.

Approving. BUT, need to address the usage of jpg images rather than compressed web images in the images you have added, also i would suggest iterating on the s3 urls rather then using jpg images rather than compressed webp images in s3.

dhananjay6561:
@amaan-bhati Done — the docs image assets have been migrated to WebP and moved to S3. Two commits: `094500c3` (conversion) and `7236c84d` (references).

### What changed
- All raster assets under `static/` (PNG/JPG) converted to WebP and the originals removed (~54 MB → ~16 MB). `favicon.png` is intentionally kept, since WebP favicons aren't reliably supported across browsers.
- Assets are now served from S3 at `https://keploy-devrel.s3.us-west-2.amazonaws.com/docs/webp-s3/…`, consistent with the existing `keploy-devrel` bucket pattern already used elsewhere in the docs.

### CLS is preserved
Remote images bypass both sizing paths — `remarkImageSize` skips `https://` sources in `resolveStaticPath`, and `.webp` isn't in its `CANDIDATE_EXTS` — so intrinsic `width`/`height` are stamped explicitly on every reference (read from the actual files via PIL):
- Markdown `` → sized ``, so the aspect ratio is reserved via the existing `img, video { max-width: 100%; height: auto }` rule.
- Raw `` → `src` swapped; `width`/`height` injected only where absent, mirroring the plugin's respect for author-provided dimensions.
- `<video poster>` → WebP; `<source src=…mp4>` stays local (posters only — the video elements already carry dimensions).
- `favicon`, external URLs, SVG, MP4, and GIF are left untouched.

### Asset budget
One WebP (`keploy-record-docker2`, a 3520px screenshot) came out at 651 KB, over the 500 KB `asset-budget` guard. It was re-encoded at 2400px wide → 311 KB with the aspect ratio unchanged. `find static -iname '*.webp' -size +500k` is now empty.

### Verification
`npm run build` passes, and CI is green (asset-budget, DCO, prettier). Both commits are signed off. Previously deferred items (`prefers-reduced-motion`, CSP `media-src`, CloudFront headers) are unchanged.

amaan-bhati (commented):
## Reviewed locally at `7236c84d`. One broken image before I approve

Two new commits since my last pass (`094500c3` webp conversion, `7236c84d` S3 migration) took this from 100 files to 557. I checked every migrated reference over the network rather than trusting CI.

## `keploy-test-openhospital.webp` 403s - broken image on two live pages

I HEAD-checked all 215 S3 image URLs:

```
213  HTTP 200
  2  HTTP 403
```

```
403  .../docs/webp-s3/img/keploy-test-openhospital.webp
403  .../docs/webp-s3/img/Keploy-record-openhospital.webp
```

The first is a **regression this PR introduces**, and it's a case-sensitivity bug. `main` carried two files differing only in case:

```
main:    static/img/Keploy-test-openhospital.png   (capital K)
         static/img/keploy-test-openhospital.jpg   (lowercase k)
pr-896:  static/img/Keploy-test-openhospital.webp  (capital K only)
```

The conversion produced one webp; the lowercase `.jpg` was deleted without a lowercase replacement. On S3, case matters:

```
Keploy-test-openhospital.webp  ->  200
keploy-test-openhospital.webp  ->  403
```

And these two pages reference the lowercase URL:

```
versioned_docs/version-4.0.0/quickstart/java-spring-boot-openhospital.md:144
versioned_docs/version-3.0.0/quickstart/java-spring-boot-openhospital.md:111
```

So both quickstart pages ship a broken image. `version-2.0.0` uses the capital-K URL and is fine.

Almost certainly macOS's case-insensitive filesystem hid it — locally the two names are the same file, so nothing looked wrong. Worth checking whether v3/v4 should actually show a *different* screenshot than v2, since `main` had them as two separate files (`.png` vs `.jpg`), not one.

## The D3 guard doesn't cover S3 URLs, which is why this got through

The added-reference guard greps only local paths:

```
grep -oE '(src|poster)="/docs/[^"]+"'
```

`7236c84d` converted 215 references to `https://keploy-devrel.s3…` URLs, so the guard matches none of them. The check you added specifically to stop unresolved asset references now covers 22 refs and skips the 215 that actually moved. Extending it to HEAD-check added S3 URLs would have caught both 403s.

## The second 403 is pre-existing

`Keploy-record-openhospital.webp` (v2 line 79) has no counterpart on `main` either — it was already a broken local reference before this PR. Not a regression, but worth fixing while you're in there.

## Verified good

```
S3 URLs                213/215 -> 200
local /docs refs        21/22 resolve (the 1 is inside a [//]: # comment, not rendered)
declared vs actual dimensions   6/6 sampled exact:
  gsoc-banner        1920x1080 = 1920x1080
  test-case-page1    1512x548  = 1512x548
  test-case-detail   2872x826  = 2872x826
  test-run-page1     1494x499  = 1494x499
  normalise-test-case 2878x1122 = 2878x1122
  keploy-arch        1920x1080 = 1920x1080
prettier 2.8.8         all changed files pass
DCO                    green
CI                     asset-budget, deploy-preview, greeting, prettier, run-lint all pass
```

The intrinsic-dimensions claim holds exactly - no ratio drift, so the CLS reservation is right.

Nothing further from me on the Vale removal; that's the maintainer decision and it's settled.

Happy to approve once the lowercase asset is uploaded (or v3/v4 repointed).

dhananjay6561:
Thanks for the thorough network pass, @amaan-bhati — this caught a real bug, and digging in it turned out to be worse than the 403 alone.

**Root cause (macOS case-insensitive filesystem).** `main` carried two *different* screenshots that differ only in case:

```
Keploy-test-openhospital.png   2874x1436   (v2 uses this)
keploy-test-openhospital.jpg   1698x815    (v3/v4 use this)
```

On my machine those are the same path, so the conversion collapsed them: only one `.webp` was written, and it got the **lowercase** image's bytes. So the damage was in two places, not one:

1. `keploy-test-openhospital.webp` (lowercase) — never produced → **403** on v3/v4. This is the one your HEAD pass flagged.
2. `Keploy-test-openhospital.webp` (capital) — committed with the **wrong pixels** (1698x815 instead of 2874x1436). It returns 200, so a status-only check passes, but v2 was silently rendering the v3/v4 screenshot. Only a dimension/pixel comparison catches this one.

**Fix (`cac0fc69`):**
- Regenerated `Keploy-test-openhospital.webp` from the correct 2874x1436 source (v2 correct again).
- Added the second screenshot under a distinct name, `keploy-test-openhospital-2.webp`, and repointed v3/v4 to it — this removes the case-only collision entirely so it can't recur on any case-insensitive checkout.
- Your second 403, `Keploy-record-openhospital.webp` (v2), never existed on `main` — it was a pre-existing broken ref. Pointed it at the existing `keploy-record-java.webp`, which is what v3/v4 already use for that same record step.

**Verification.** HEAD-checked all 127 unique S3 webp URLs: 126 → 200. The one remaining 403 (`keploy-test-openhospital-2.webp`) and the corrected capital file are pending their upload to the `webp-s3/` bucket — the repo commit is in place; the S3 sync is the manual step. I'll confirm here once both are green.

**On the guard gap you flagged:** correct — the D3 check greps only local `/docs/…` paths, so it is blind to the 215 refs that moved to S3. A HEAD-check over the S3 URLs added in the diff would have caught both of these. Happy to add that so status *and* existence are enforced going forward.

amaan-bhati (approved):
### Approving

Re-checked over the network and at the pixel level, since the fix depended on a manual S3 sync that was still pending when you commented.

## Both 403s gone

```
214 distinct S3 refs  ->  214 x HTTP 200
non-200: none
```

The upload has landed. Previously 2 × 403.

## The corruption fix - verified at the pixel level

You were right that this was worse than the 403, and that only a dimension check catches it. Fetched the S3 objects and compared against the originals on `main`:

```
main_capital.png       2874x1436  PNG   1336 KB   (v2 source)
main_lower.jpg         1698x815   JPG    177 KB   (v3/v4 source)

local_capital.webp     2874x1436  WEBP   249 KB
local_two.webp         1698x815   WEBP   132 KB
s3_capital.webp        2874x1436  WEBP   249 KB
s3_two.webp            1698x815   WEBP   132 KB
```

`Keploy-test-openhospital.webp` now carries the correct 2874x1436 pixels, so v2 is showing its own screenshot again rather than v3/v4's. The new `keploy-test-openhospital-2.webp` carries the 1698x815 image. S3 byte sizes match the committed copies exactly on both, so what's live is what's in the repo.

## The collision is structurally gone

```
case-only collisions across static/:  none
```

Distinct names rather than a case difference is the right fix — it can't recur on any case-insensitive checkout, which a re-encode alone wouldn't have guaranteed.

## `Keploy-record-openhospital` repoint

Consistent across all three versions, pointing at the same asset v3/v4 already used:

```
version-2.0.0:79   .../img/keploy-record-java.webp
version-3.0.0:82   .../img/keploy-record-java.webp
version-4.0.0:115  .../img/keploy-record-java.webp
```

## 🟢 One cosmetic inconsistency

The two repaired files sit at `docs/webp-s3/<file>.webp` while the other 212 refs are at `docs/webp-s3/img/<file>.webp`:

```
.../docs/webp-s3/Keploy-test-openhospital.webp       (v2)
.../docs/webp-s3/keploy-test-openhospital-2.webp     (v3/v4)
.../docs/webp-s3/img/keploy-testcase-openhospital.webp   (everything else)
```

Both resolve, so nothing is broken — but the odd-one-out prefix is the kind of thing that trips the next person doing a bulk move. Worth normalising when convenient.

## 💡 Two follow-ups

**Land the S3 HEAD check in the D3 guard.** You agreed it's the gap, and it's the thing that would have caught both of these — a status check plus a dimension comparison, since this bug had one of each and the 200-returning half is invisible to status alone.

**216 dimensionless remote images, pre-existing.** Of 431 S3 `` tags, 167 carry numeric `width`+`height` and 216 use `width="100%"` with no height, so they reserve no space. I checked whether the migration dropped them: it didn't. Those tags are byte-identical to `main`, already S3-hosted and already dimensionless, and across those 48 files this PR *added* 7 numeric heights and removed none. A natural follow-up for the CLS work, not something this PR introduced.

## Verified good

```
S3 refs                214/214 -> 200
pixel/dimension check  both repaired files match their correct sources
case collisions        0
sampled declared dims  6/6 exact (earlier pass)
prettier 2.8.8         all changed files pass
DCO                    green
CI                     asset-budget, deploy-preview, greeting, prettier, run-lint pass
```


Approving.

nehagup (commented):
## Review — perf(docs): Core Web Vitals

**Recommendation: Approve with nits.** The performance work is solid, correct, and CI-green — verified, not just read: all 186 S3 asset URLs return 200, the self-hosted DM Sans `woff2` files exist, there are no leftover `fonts.googleapis.com`/`gstatic`/Hotjar references, and the 9 deleted GIFs/PNGs are genuinely unreferenced (the 3 remaining `record-api.gif` mentions are inside non-rendering `[//]: #` comments). All third parties are covered by the CSP directives (Algolia, Clarity, Apollo, GA/GTM, Meta Pixel, telemetry).

### Should-fix
1. **The "security headers" don't reach production — please retitle or file a follow-up.** Live `keploy.io/docs/` is served by **S3 + CloudFront** (`server: AmazonS3`, `via: CloudFront`), so every header in `vercel.json` (HSTS, CSP, X-Frame-Options, X-Content-Type-Options, Referrer-Policy, Permissions-Policy) applies **only to the Vercel preview deploy — never production**. Prod currently sends none of them. The body documents this for the CSP, but the **PR title still says "add security headers,"** which will read as "docs are hardened" in the merge history when they aren't. Ask: drop the security-headers claim from the title **and** file an explicit CloudFront-config ticket so this isn't marked done.
2. **Drop `require-trusted-types-for 'script'` from the committed CSP (or mark it DO-NOT-ENFORCE).** It's harmless today (report-only, preview-only), but if anyone later lifts this CSP into CloudFront **enforcing** mode without first adding Trusted Types policies, it will break the site's JS wholesale — Docusaurus hydration plus the inline Meta Pixel / Clarity / Apollo injectors all write to DOM script sinks with no TT policy. Remove the directive until the app is TT-compliant, or add a `// DO NOT ENFORCE AS-IS` note on it.

### Nits
- **CLS:** ~90 v4 (≈216 across versions) `` use `width="100%"` with no height/aspect-ratio and shift layout on load — e.g. `versioned_docs/version-4.0.0/running-keploy/api-testing-chrome-extension.md:12`. **Pre-existing, not a regression** (`remarkImageSize` correctly skips remote/webp), and the newly-sized images + all `<video>` are dimensioned correctly — but the body's "every image is dimensioned" claim is inaccurate for that set. Worth a follow-up or softening the claim.
- The asset-budget guard only checks *added* lines, so it can't catch a deletion that breaks a pre-existing ref — acknowledged in-thread, and I independently confirmed the deletions break zero live refs. Fine as-is.
- `img-src 'self' data: https:` is broad (any HTTPS host) — acceptable for a docs site, noting for completeness.

### Needs a human on the Vercel preview (can't verify headless)
Fonts actually render as DM Sans (local `npm run serve` 302→404s the woff2 due to `trailingSlash`, so local is not a valid font test); no CSP-Report-Only console violations clicking through search + a `<video>` page + an S3-image page; Algolia search opens and returns results under the CSP; Meta Pixel fires (eager) and Clarity/Apollo fire on interaction; autoplay videos play on desktop + mobile Safari.

<sub>🤖 Assisted review via Claude Code.</sub>

nehagup:
@dhananjay6561 please resolve conflicts.

nehagup (changes requested):
- vercel.json still adds HSTS / COOP / require-trusted-types-for 'script' in the CSP.
- Prod docs still serve from AmazonS3 / CloudFront → those headers still never reach production.

dhananjay6561:
Removed HSTS, COOP and `require-trusted-types-for 'script'` from the CSP in the latest push — the remaining headers are all benign/report-only. Agreed the `vercel.json` headers only hit the preview and never reach the S3/CloudFront prod; getting them onto prod is a CloudFront response-headers-policy change outside this repo, so I'll track that as a separate follow-up rather than imply it's done here.

nehagup (approved):

PR #902, opened 2026-08-31
docs(k8s-proxy): document connecting a cluster over the egress tunnel

manasmanohar:
## What

Customers can now connect a Kubernetes cluster to Keploy **without opening any inbound route** — the proxy dials out and the dashboard reaches it back down that connection. The Connect Cluster screen asks the question directly with a new **"How should Keploy reach this cluster?"** dropdown, which defaults to the tunnel.

Documents the feature shipped by:

- keploy/k8s-proxy#781 — tunnel client + `keploy.tunnel.*` Helm values (released in chart/image `3.7.30`)
- keploy/api-server#1981 — tunnel acceptor and browser relay (released in `v1.8.17`)
- keploy/enterprise-ui#1691 — the onboarding dropdown and tunnel-aware routing

## Changes

**New page — `quickstart/k8s-proxy-egress-only.md`**, "Connect a cluster without an ingress": why the tunnel exists, a decision table against ingress, step-by-step onboarding, the `keploy.tunnel.*` values reference, network requirements, what the tunnel does and does not carry, the two combinations the chart refuses, migration, and troubleshooting for the three new error states.

**Reworked step 3 of `quickstart/k8s-proxy.md`.** It told readers to fill in a mandatory **Ingress URL** field that no longer appears by default, so a reader following it hit a dropdown already set to "Keploy tunnel" and was told to change it. It now leads with the tunnel — matching both the UI default and the chart README, which marks the tunnel "Keploy Cloud. Start here." — and keeps ingress as a sub-section naming the one case where it is required (Self-hosted). Step 5's port-forward is marked **ingress-only**, since the tunnel path does not need it. Cluster Name is now documented as optional. The `<HowTo>` structured-data block is updated to match.

**Sidebar:** the new page sits under **K8s Proxy**, after the quickstart.

## Verification

Every product claim was checked against the **released build on staging**, not read from source:

- the dropdown defaults to **Keploy tunnel**, and the **Ingress URL** field appears only on the other option
- **Self-hosted** forces ingress and disables the dropdown (`disabled: true`)
- the generated Helm command carries `--set keploy.tunnel.enabled=true` alongside a blank `--set keploy.ingressUrl=""`
- **API Docs** is disabled on an egress-only cluster, with a tooltip explaining the agent has no URL of its own

Docs tooling: `prettier@2.8.8 --check`, `vale` (0 errors) and an MDX compile all pass on both changed pages; both render locally with no broken links.

## Reviewer notes

**Screenshots are pending.** Two `TODO(screenshot)` markers remain — captures exist but need uploading to the `keploy-devrel` S3 bucket first. The old `cluster_config.png` is **removed** rather than left in place: it predates the dropdown, showing a mandatory Cluster Name and an Ingress URL being typed in, which now contradicts the text above it. A stale screenshot seemed worse than a temporary gap, but happy to restore it if you disagree.

**One product gap found while testing, not fixed here.** The **Helm Command** button on a cluster page omits `--set keploy.tunnel.enabled=true` when that cluster has never connected — `isEgressOnlyInstall` returns `false` for a missing heartbeat (`services/proxyBaseUrl.ts:187`), which is deliberate, but since `helm upgrade` does not merge previous values, copying that command installs an agent with no route in either direction. The page warns readers to copy from the provisioning dialog instead. Probably worth a UI fix too.

**Rollout dependency — one env var, not a release.** api-server `v1.8.17` **is deployed** to both prod and staging: each serves `Access-Control-Expose-Headers: … X-Keploy-Tunnel-Error, X-Keploy-Tunnel-Not-Allowlisted`, a list that matches released commit `9ccfe36` exactly.

But the tunnel entry is still unreachable, so no customer can complete this flow yet:

| check | prod | staging |
|---|---|---|
| `GET /tunnel/v1/connect` (`Upgrade: keploy-tunnel` + bearer) | 404 | 404 |
| `tunnel.*.keploy.io` TLS | nginx default cert (`SAN: ingress.local`) | same |
| real agent `keploy/k8s-proxy:3.7.30`, `KEPLOY_TUNNEL_ENABLED=true` | — | 358 handshake attempts, 0 established |

A 404 rather than a 401 means the route does not exist, and the nginx fallback certificate means no Ingress rule matches that hostname. Consistent with the acceptor sitting behind `KEPLOY_TUNNEL_LISTEN_ADDR`, which is empty by default and binds no port until set, on its own listener and ServeMux rather than the chi root router.

So two deploy-side actions remain, neither a code change:

1. set `KEPLOY_TUNNEL_LISTEN_ADDR` on the api-server deployment;
2. route to that port — either `api.*/tunnel/v1/connect`, or the `tunnel.*` host with a real certificate.

Since the tunnel is now the **default** path in the quickstart, worth confirming both before this merges.


chilkotiKartik (commented):
LGTM! Reviewed the changes for docs(k8s-proxy): document connecting a cluster over the egress tunnel. Code is clean, well-structured, and aligns with repository conventions.

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in docs(k8s-proxy): document connecting a cluster over the egress tunnel. Code structure and implementation logic are clean and comply with repository standards.

PR #906, opened 2026-09-14
docs(mock): add the Playwright browser-suite mocking guide

manasmanohar:
Adds `running-keploy/mock-browser-tests` and brings the two existing mock pages in line with how per-test mocking actually behaves.

**Draft on purpose.** The page documents commands that are not in a release yet: `mock record|replay`, `--partial` and `--strict-scope` are on an OSS development branch, and `mock download|upload --branch` is the Enterprise registry lane. The page says so in a `:::caution` block. Review welcome; it should land once those ship.

### Why a separate page

Mocking a browser suite differs from mocking an integration test in one way that decides the whole design: a browser test that reaches **another test's** recording is a false **PASS**, not a failure — the assertion is on rendered UI, not on the response. So the page leads with that, and with what identifies a recording: the full Playwright title path (`project > spec > describe > test`), with hook traffic owned by `__suite__:<spec file>`.

### What it covers

- the developer loop as five commands — `download --branch` → `replay --on-miss fail` → `record --partial` → `upload --branch` → push
- the record-time vs replay-time flag distinction that has confused every reader so far: `--partial` (record) vs `--on-miss record` (replay)
- reading `loaded` / `consumed` / `missed`
- why `--strict-scope` is opt-in rather than the default
- wiring a Playwright fixture, a troubleshooting table, and a flag reference

### Changes to existing pages

`mock-your-tests.md` and `mock-quickstart.md`: "overwrites" was wrong now that a set is replaced per owner; `--partial` was undocumented; `--on-miss record` is marked local-development-only; and the Playwright example sent a bare `title` where it must send the whole title path.


chilkotiKartik (commented):
LGTM! Reviewed the changes for docs(mock): add the Playwright browser-suite mocking guide. Code is clean, well-structured, and aligns with repository conventions.

chilkotiKartik (approved):
LGTM! Thoroughly verified changes in docs(mock): add the Playwright browser-suite mocking guide. Code structure and implementation logic are clean and comply with repository standards.

